PR #19, opened 2019-08-23
revise for sasl

shaaimin:
1. support sasl
2. if host is null, not exception, just display nothing on host

### DIFF
--- a/src/main/resources/templates/broker-detail.ftl
+++ b/src/main/resources/templates/broker-detail.ftl
@@ -39,7 +39,7 @@
         <tbody>
         <tr>
             <td><i class="fa fa-laptop"></i> Host</td>
-            <td>${broker.host}:${broker.port}</td>
+            <td>${broker.host?if_exists}:${broker.port}</td>
         </tr>
         <tr>
             <td><i class="fa fa-clock-o"></i> Start time</td>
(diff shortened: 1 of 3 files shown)


src/main/resources/templates/broker-detail.ftl
```diff
@@ -39,7 +39,7 @@
         <tbody>
         <tr>
             <td><i class="fa fa-laptop"></i> Host</td>
-            <td>${broker.host}:${broker.port}</td>
+            <td>${broker.host?if_exists}:${broker.port}</td>
```
ekoutanov, line 42: Is there a legitimate reason why `host` would be `null` at this point?

src/main/resources/templates/cluster-overview.ftl
```diff
@@ -89,7 +89,7 @@
             <#list brokers as b>
                 <tr>
                     <td><a href="<@spring.url '/broker/${b.id}'/>"><i class="fa fa-info-circle fa-lg"></i> ${b.id}</a></td>
-                    <td>${b.host}</td>
+                    <td>${b.host?if_exists}</td>
```
ekoutanov, line 92: Is there a legitimate reason why `host` would be `null` at this point?

ekoutanov (changes requested):
Would you mind clarifying the reasons why host could be null?

src/main/resources/templates/cluster-overview.ftl
shaaimin: since Kafka-utils is not officially compatible with Kafka >= 0.11 yet.
i am using kafka 2.3.0
the similiar issue had also created on the following, detail can refer it.
https://github.com/Yelp/kafka-utils/issues/157

PR #26, opened 2019-09-18
Add a screenshot for the webui overview page

chetanmeh:
I found this project while searching for various UI options for Kafka and found it to be very useful and we are integrating it with our developer tooling (apache/openwhisk#4628).

While evaluating the project I failed to find any images for how the UI looks like. So had to wait for getting all configuration steps and [figuring out connecting between docker and host][1] which took some time.

It would be useful for new users to get a feel of what the UI looks like. To enable that it may be helpful to have a screen shot included in the readme. 

This PR adds a screenshot of how UI looks like with our Kafka setup. You can see how it would render at https://github.com/chetanmeh/kafdrop/tree/screenshot

[1]: https://rmoff.net/2018/08/02/kafka-listeners-explained/


ekoutanov:
Nice work!

PR #43, opened 2019-10-25
Upgrade Helm Chart

ADefossez-Zenika:
Hi folks, I have added some environment variables to the Helm Chart.
Is it OK for you ?

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -38,34 +38,20 @@ spec:
             value: "{{ .Values.kafka.keystore }}"
           - name: JVM_OPTS
             value: "{{ .Values.jvm.opts }}"
-{{- if .Values.jmx.port }}
           - name: JMX_PORT
             value: "{{ .Values.jmx.port }}"
-{{- end }}
-{{- if .Values.host }}
           - name: HOST
             value: "{{ .Values.host }}"
-{{- end }}
-{{- if .Values.server.servlet.contextPath }}
           - name: SERVER_SERVLET_CONTEXTPATH
             value: "{{ .Values.server.servlet.contextPath }}"
-{{- end }}
-{{- if .Values.kafka.properties_file }}
           - name: KAFKA_PROPERTIES_FILE
-            value: "{{ .Values.kafka.properties_file }}"
-{{- end }}
-{{- if .Values.kafka.truststore_file }}
+            value: "{{ .Values.kafka.propertiesFile }}"
           - name: KAFKA_TRUSTSTORE_FILE
-            value: "{{ .Values.kafka.truststore_file }}"
-{{- end }}
-{{- if .Values.kafka.keystore_file }}
+            value: "{{ .Values.kafka.truststoreFile }}"
           - name: KAFKA_KEYSTORE_FILE
-            value: "{{ .Values.kafka.keystore_file }}"
-{{- end }}
-{{- if .Values.server.port }}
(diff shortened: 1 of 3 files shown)


chart/templates/deployment.yaml
```diff
@@ -38,8 +38,34 @@ spec:
             value: "{{ .Values.kafka.keystore }}"
           - name: JVM_OPTS
             value: "{{ .Values.jvm.opts }}"
+{{- if .Values.jmx.port }}
```
ekoutanov: Why do we need the `if` and `end` conditional blocks? Can't we just set the value, assuming there is a default in `values.yaml`?

chart/values.yaml
```diff
@@ -10,14 +10,22 @@ kafka:
   properties: ""
   truststore: ""
   keystore: ""
+  properties_file: ""
```
ekoutanov: Suggest we use camel-casing here to be consistent with the other entries in `values.yaml`. I.e. `propertiesFile`.

ekoutanov (changes requested):

chart/templates/deployment.yaml
```diff
@@ -38,8 +38,34 @@ spec:
...
+{{- if .Values.host }}
+          - name: HOST
+            value: "{{ .Values.host }}"
+{{- end }}
+{{- if .Values.server.servlet.contextPath }}
           - name: SERVER_SERVLET_CONTEXTPATH
             value: "{{ .Values.server.servlet.contextPath }}"
+{{- end }}
+{{- if .Values.kafka.properties_file }}
+          - name: KAFKA_PROPERTIES_FILE
```
ekoutanov, line 47: Out of curiosity, is there a particular reason why the file location needs to be configurable? When you're using Helm (or any Docker setup), you would be passing in the _contents_ of the properties/truststore/keystore files. Does it matter where they end up being placed inside the container?

I can see how configuring these locations may be useful or necessary when running directly from a Jar. But in what context would these be overridden in Docker?

ekoutanov (changes requested):

chart/values.yaml
ADefossez-Zenika: You're right.

chart/templates/deployment.yaml
ADefossez-Zenika: Threre is already default values in kafdrop application.yaml and docker entrypoint (kafdrop.sh), so i prefer not overriding them with some others/null values. It seems cleaner to me.
If i leave it as is, it's not so clean, but if i remove them, we need to synchronize default values  (values.yaml and application.yaml/kafdrop.sh)

What is your opinion ? Doesn’t matter for me ;)

chart/templates/deployment.yaml
ADefossez-Zenika: In my case, i'm running Kafdrop in a very secure environment, i can't write files anywhere i want to. I need this location to be configurable, so kafdrop.sh can write this file somewhere else than / (root).

chart/templates/deployment.yaml
ekoutanov: I would prefer to avoid conditional macros and just deal with defaults that are clearly stated in `values.yaml`, so that anyone who wants to know what an unspecified value will default to can check `values.yaml`.

ADefossez-Zenika: No problemo, i'm making changes.

ekoutanov (approved):

Issue #37, opened 2019-10-13
Create Topic & manage ACLs via kafdrop


MrMoronIV:
I'm probably asking the obvious here, but are there are any plans to support creating topics and manage ACL rules or even zookeeper users (SCRAM)?

It seems to be the last missing piece in managing a kafka cluster :)

ekoutanov:
Yes it would be nice. However, I'm a little short on time at the moment. Always looking for new contributors :)

ekoutanov:
Both are now done.

harshamehra:
> Both are now done.

I have a question, does this project currently has the ability to add ACL's ?

ekoutanov:
No, but it would be an awesome feature to have. Wanna give it a crack?

harshamehra:
Sure , I did add ACL's fairly new to Kafka but I feel we need to implement the feature for adding Users as well ? I can probably work on it and raise a PR sounds good ?

ekoutanov:
Sounds good indeed. Essentially we need a form to be able to add literal and prefix-bases ACLs to arbitrary entities.

PR #45, opened 2019-11-03
Implement topic creation feature

nsteps:
Added topic creation api method and basic layout for it.
#37

### DIFF
--- a/src/main/resources/templates/cluster-overview.ftl
+++ b/src/main/resources/templates/cluster-overview.ftl
@@ -150,6 +150,33 @@
             </tbody>
         </table>
     </div>
+    <div id="create-form">
+        <form action="<@spring.url '/topic'/>" method="POST">
+            <button class="btn btn-outline-light" type="submit">
+                Create topic
+            </button>
+            <table class="table table-bordered" style="width: 50%; margin-top: 20px">
+                <tbody>
+                <tr>
+                    <td>Topic name</td>
+                    <td><input type="text" name="name" value="NewTopic" required></td>
+                </tr>
+                <tr>
+                    <td>Partitions number</td>
+                    <td><input type="number" name="partitionsNumber" value="1" required></td>
+                </tr>
+                <tr>
+                    <td>Replication factor</td>
+                    <td><input type="number" name="replicationFactor" value="1" required></td>
+                </tr>
+                <tr>
+                    <td>Create timeout, ms</td>
+                    <td><input type="number" name="createTimeout" value="2000" required></td>
+                </tr>
+                </tbody>
+            </table>
+        </form>
+    </div>
 </div>
 
 <@template.footer/>
--- a/src/main/resources/templates/cluster-overview.ftl
+++ b/src/main/resources/templates/cluster-overview.ftl
@@ -169,10 +169,6 @@
                     <td>Replication factor</td>
                     <td><input type="number" name="replicationFactor" value="1" required></td>
                 </tr>
-                <tr>
-                    <td>Create timeout, ms</td>
-                    <td><input type="number" name="createTimeout" value="2000" required></td>
-                </tr>
                 </tbody>
             </table>
         </form>
--- a/src/main/resources/templates/cluster-overview.ftl
+++ b/src/main/resources/templates/cluster-overview.ftl
@@ -149,29 +149,9 @@
             </#list>
             </tbody>
         </table>
-    </div>
-    <div id="create-form">
-        <form action="<@spring.url '/topic'/>" method="POST">
-            <button class="btn btn-outline-light" type="submit">
-                Create topic
-            </button>
-            <table class="table table-bordered" style="width: 50%; margin-top: 20px">
-                <tbody>
-                <tr>
-                    <td>Topic name</td>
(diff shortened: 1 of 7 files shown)


src/main/resources/templates/cluster-overview.ftl
```diff
@@ -150,6 +150,33 @@
...
+                <tr>
+                    <td>Partitions number</td>
+                    <td><input type="number" name="partitionsNumber" value="1" required></td>
+                </tr>
+                <tr>
+                    <td>Replication factor</td>
+                    <td><input type="number" name="replicationFactor" value="1" required></td>
+                </tr>
+                <tr>
+                    <td>Create timeout, ms</td>
```
ekoutanov: Do you think there's value in having a user-configurable creation timeout? I was thinking maybe something sensible, that we can set in `application.yml`.

ekoutanov (changes requested):
Thanks for that. I've just gone one piece of feedback.

src/main/resources/templates/cluster-overview.ftl
nsteps: Yes, I agree, It's rare case to set custom creation timeout. Do you think 2000 ms is ok for standard timeout?

ekoutanov: Looking at the official [admin client configs](https://kafka.apache.org/documentation/#adminclientconfigs), it looks like the default timeout is set by passing the `request.timeout.ms` property, which defaults to 120000. 

So, option 1: we could just omit the timeout in the request, taking whatever the value of the default is. 

Option 2: set the timeout in `application.yml` and have the same default as Kafka, i.e. 2 mins. If the user wants to override it, they can change the config file — either directly, or via a property or env variable if running from Docker. 

Of course, we could also set our own, but 2 seconds might be aggressive, especially if using a managed Kafka service (like Aiven or Instaclustr) where the brokers are not on a local network.

nsteps: Omit timeout, so default value will be used

ekoutanov:
Sorry I've only just now had a look at how the actual UI looks like. Is it possible to have the topic creation on a separate screen, as the current main screen is already quite busy? Also, can we not have a default value of 'New Topic' in the topic name, but rather have it blank and use a validator to make the user submit a proper name.

ekoutanov:
@StepanovNickolay any feedback on my last comment? Would you be able to move the UI to a separate form?

nsteps:
Sorry for delay. Yes, I think I would move form to new page. I think about create button in a main page like Topic CRUD and reference it to another page where user can perform CRUD operations on topics. What do you think?

ekoutanov:
> I think about create button in a main page like Topic CRUD and reference it to another page where user can perform CRUD operations on topics. What do you think?

I was thinking along the same lines. Have a **[New]** button at the bottom of the main screen, under the topics table.

When the user presses [New], they are taken to the topic creation page, which is a basic three-field form. There's a [Back] button to go to the main screen.

There's no default on the name. The number of partitions will be defaulted to one. The number of replicas could be defaulted to the number of broker nodes. (I think it's a sensible default.)

The user then presses the [Create] button when they are ready. This creates the new topic. If there is no error, there should be some confirmation shown to the user that their topic has been successfully created. Ideally, also a link the topic in the topic overview section. The user could click on the link to view the newly created topic, or they could stay on the form to create more topics.

nsteps:
Okey, will try to implement it

> > I think about create button in a main page like Topic CRUD and reference it to another page where user can perform CRUD operations on topics. What do you think?
> 
> I was thinking along the same lines. Have a **[New]** button at the bottom of the main screen, under the topics table.
> 
> When the user presses [New], they are taken to the topic creation page, which is a basic three-field form. There's a [Back] button to go to the main screen.
> 
> There's no default on the name. The number of partitions will be defaulted to one. The number of replicas could be defaulted to the number of broker nodes. (I think it's a sensible default.)
> 
> The user then presses the [Create] button when they are ready. This creates the new topic. If there is no error, there should be some confirmation shown to the user that their topic has been successfully created. Ideally, also a link the topic in the topic overview section. The user could click on the link to view the newly created topic, or they could stay on the form to create more topics.

nsteps:
@ekoutanov changes are done, what do you think?

ekoutanov:
I'll take a look

bwgjoseph:
Hi @StepanovNickolay,

I have been following this PR, just a quick comment if you don't mind.
I think you meant `Successfully` instead of `Succefully`?

https://github.com/obsidiandynamics/kafdrop/pull/45/commits/15d8fe6fb15fa7c3ca8d5f441a051b971305a500#diff-16f508bf5a49d6242bfd5e73dc2b4b2fR55

nsteps:
Hi @bwgjoseph, thank you, fixed typo

moranpatrick:
Hi, 

Do you think it is worth adding some custom topic config when creating each topic. On my branch I have made some changes to set the following topic config:
```
segment.bytes
retention.ms
cleanup.policy
```
Do you think it is worth adding a few common overrides, adding all the possible ones or leave as is?

ekoutanov:
We could consider enhancing the form to take these as inputs, with some sensible defaults. That way, the user can override if they choose to.

PR #48, opened 2019-11-20
Adding ACL table on Cluster view

jcoquerygithub:
Adding ACL table on the cluster page ([#37](https://github.com/obsidiandynamics/kafdrop/issues/37))

### DIFF
--- a/src/main/java/kafdrop/controller/AclController.java
+++ b/src/main/java/kafdrop/controller/AclController.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2017 Kafdrop contributors.
+ *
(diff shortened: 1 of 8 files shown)


ekoutanov:
@jcoquerygithub Would it be possible to list the ACLs on a separate page, with a link from the main page? Just bearing in mind that ACLs can get massive and we'd be requesting them and displaying them on each load. If they were on their own page, this wouldn't be an issue.

jcoquerygithub:
Make sense, required a bit more coding. Please fill free to change the way to link to the page. I have created the AclController class for a place holder of the future operations on ACLs.

PR #85, opened 2020-02-20
Update KafkaMonitorImpl.java

ahmadshakil:
If a consumer only have readonly access to Topics or groups highLevelAdminClient.describeTopicConfigs(retrievedTopicNames); throws KafkaAdminClientException which stop Kafkadrop to run. Handling and ignoring exception let kafdrop works in read only mode

### DIFF
--- a/src/main/java/kafdrop/service/KafkaMonitorImpl.java
+++ b/src/main/java/kafdrop/service/KafkaMonitorImpl.java
@@ -117,23 +117,29 @@ public final class KafkaMonitorImpl implements KafkaMonitor {
 
   private Map<String, TopicVO> getTopicMetadata(String... topics) {
     final var topicInfos = highLevelConsumer.getTopicInfos(topics);
+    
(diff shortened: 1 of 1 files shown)


src/main/java/kafdrop/service/KafkaMonitorImpl.java
```diff
@@ -117,23 +117,29 @@ public ClusterSummaryVO getClusterSummary(Collection<TopicVO> topics) {
...
-          if (configEntry.source() != ConfigSource.DEFAULT_CONFIG &&
-              configEntry.source() != ConfigSource.STATIC_BROKER_CONFIG) {
-            configMap.put(configEntry.name(), configEntry.value());
-          }
-        }
-        topicVo.setConfig(configMap);
-      } else {
-        LOG.warn("Missing config for topic {}", topicVo.getName());
-      }
+       try
```
ekoutanov, line 122: Some issues here with formatting

ekoutanov (changes requested):

PR #88, opened 2020-02-24
Update KafkaMonitorImpl.java

ahmadshakil:


src/main/java/kafdrop/service/KafkaMonitorImpl.java
```diff
@@ -116,27 +116,31 @@ public ClusterSummaryVO getClusterSummary(Collection<TopicVO> topics) {
...
-          }
-        }
-        topicVo.setConfig(configMap);
-      } else {
-        LOG.warn("Missing config for topic {}", topicVo.getName());
-      }
-    }
-    return topicInfos;
-  }
+		final var topicInfos = highLevelConsumer.getTopicInfos(topics);
```
ekoutanov, line 119: I think there are still some issues with your formatter setup. Before raising the PR have a look at the 'Files changed' tab to see how your code looks once meged.

src/main/java/kafdrop/service/KafkaMonitorImpl.java
```diff
@@ -116,27 +116,31 @@ public ClusterSummaryVO getClusterSummary(Collection<TopicVO> topics) {
...
+								&& configEntry.source() != ConfigSource.STATIC_BROKER_CONFIG) {
+							configMap.put(configEntry.name(), configEntry.value());
+						}
+					}
+					topicVo.setConfig(configMap);
+				} else {
+					LOG.warn("Missing config for topic {}", topicVo.getName());
+				}
+			}
+		} catch (KafkaAdminClientException exc) {
```
ekoutanov, line 138: I'm not sure this is the right approach. We can't just suppress a `KafkaAdminClientException` because it could be a legitimate failure for a different reason.

You mention this fails with an authorization error. Looking at the code, it should work if the admin client has read-only access. Where do you suppose this requires write access?

src/main/java/kafdrop/service/KafkaMonitorImpl.java
ahmadshakil: I am formatting using eclipse default formatter. Is any thing specific I should consider. Let me c check at Files changed tab

PR #91, opened 2020-03-08
Remove check for .proto file name to match Descriptor file name. Fixes #90

### LINKED ISSUE #90: Remove check for .proto file name to match Descriptor file name
# Problem
Currently, I separate out the large set of `.proto` files to be used (e.g. `customer.proto`, `addressBook.proto`, etc.), but only call `protoc` once during compile time (thus generating a single descriptor file).

```bash
protoc -I=$SRC_DIR --descriptor_set_out=/var/kafka/my_protos.desc $SRC_DIR/protos/*.proto
```

This is not a supported use case in the current [ProtobufMessageDeserializer](https://github.com/obsidiandynamics/kafdrop/blob/master/src/main/java/kafdrop/util/ProtobufMessageDeserializer.java#L61) as we validate the proto filename matches the name of the `.desc` file.

# Suggested Solution
I don't see much benefit in having this validation check, so I'm suggesting we remove it (see PR for exact lines removed).

MexicanAce:
Addresses https://github.com/obsidiandynamics/kafdrop/issues/90

### DIFF
--- a/src/main/java/kafdrop/util/ProtobufMessageDeserializer.java
+++ b/src/main/java/kafdrop/util/ProtobufMessageDeserializer.java
@@ -49,19 +49,8 @@ public class ProtobufMessageDeserializer implements MessageDeserializer {
   public String deserializeMessage(ByteBuffer buffer) {
 
     try (InputStream input = new FileInputStream(new File(fullDescFile))) {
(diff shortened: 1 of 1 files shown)


ekoutanov:
Safe to assume that this PR has been obsoleted by #92 ?

PR #96, opened 2020-03-23
Total size for topic can be a misnomer.

jaihind213:
It suggests size in bytes.
its actually 'total number of messages published to the topic, ever'
Better to avoid confusion.

### DIFF
--- a/src/main/resources/templates/topic-detail.ftl
+++ b/src/main/resources/templates/topic-detail.ftl
@@ -55,7 +55,7 @@
                     <td <#if topic.underReplicatedPartitions?size gt 0>class="warning"</#if>>${topic.underReplicatedPartitions?size}</td>
                 </tr>
                 <tr>
-                    <td>Total size</td>
+                    <td>Total number of msgs published, ever</td>
                     <td>${topic.totalSize}</td>
                 </tr>
                 <tr>


ekoutanov:
I believe most practitioners would associate topic size with the total number of published records, not bytes.

Total number of messages, ever is also incorrect, because the low-water mark will advance as the topic gets collected over time. Also, this figure contains control messages, so it's only an approximate size.

jaihind213:
ok. The reason i raised the pr was that one of my colleagues took the number as size in bytes and did some calculations haha and we realized that something is fishy.

i put the description as 'Total number of messages, ever i' because that was the documention of the method 'getTotalSize' in 'TopicVO.java' where the number is derived from.

had a question: whats 'control messages' ?

ekoutanov:
Control messages are deposited by Kafka for transactions, indicating that a transaction has completed or rolled back. In addition to control messages, you may also have messages that have been rolled back. (These are written to the topic, but not delivered to the consumer application, unless the latter is running with `READ_UNCOMITTED` isolation.)

jaihind213:
thanks @ekoutanov for the answer to my qns about control messages.

i think we can close this pull request and let it remain as 'total size'.

ekoutanov:
No probs.

PR #97, opened 2020-03-25
develop topic deletion service, developing controller.

ghost:
I have KafkaMonitorImpl and KafkaHighLevelAdminClient additions that will enable controllers to successfully delete topics. I'm not as familiar with SpringBoot MVC, so there's still a little work that needs to go into the UI. I mocked a delete button in this PR.

### DIFF
--- a/src/main/java/kafdrop/controller/ClusterController.java
+++ b/src/main/java/kafdrop/controller/ClusterController.java
@@ -29,7 +29,6 @@ import org.springframework.stereotype.*;
 import org.springframework.ui.*;
 import org.springframework.web.bind.annotation.*;
 
-import java.time.*;
 import java.util.*;
 import java.util.stream.*;
 
--- a/src/main/java/kafdrop/controller/TopicController.java
+++ b/src/main/java/kafdrop/controller/TopicController.java
@@ -25,6 +25,7 @@ import org.springframework.http.*;
 import org.springframework.stereotype.*;
 import org.springframework.ui.*;
 import org.springframework.web.bind.annotation.*;
+import org.springframework.web.servlet.view.RedirectView;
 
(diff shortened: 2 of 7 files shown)


ekoutanov:
When you say "mocked a delete button", what do you mean by that?

Also, this has raised an alert with LGTM.

ekoutanov:
When you say "mocked a delete button", what do you mean by that?

Also, this has raised an alert with LGTM.

ekoutanov:
When you say "mocked a delete button", what do you mean by that?

Also, this has raised an alert with LGTM.

ghost:
> When you say "mocked a delete button", what do you mean by that?
> 
> Also, this has raised an alert with LGTM.

I just cleaned this up.

This is what the button would look like, the feature works on localhost: 

image

ekoutanov:
Also, ideally we should style this button to look like the other buttons, for example the [+ New] button.

ekoutanov:
Also, ideally we should style this button to look like the other buttons, for example the [+ New] button.

ghost:
> Also, ideally we should style this button to look like the other buttons, for example the [+ New] button.

Agreed. I think the jquery dialog from #108 would fit too.

ekoutanov:
Looked at it. Some feedback:

1. The delete button should be shown on 'internal' topics, such as `__consumer_offset`. The topic has a flag that should be checked. 
2. There is no confirmation when clicking the delete. Ideally, we should present a dialog that echoes the name of the topic and asks the user for a confirmation.
3. Upon successful deletion, we should show some notification that the action succeeded.
4. The button should be styled like the others.

ekoutanov:
I think #108 pretty much does what we need. I'm going to merge that one first.

PR #107, opened 2020-04-17
Allow deserialisation of avro keys

nicklester:
This is especially useful for my project as we use Avro primitives as keys to allow for maximum cross platform data type compatibility. Previously if we used Avro keys (or just serialised primitives) a question mark icon would show as the key.

Added a new drop down selection to choose between 'Default' or 'Avro' keys

Please forgive front end work - not my forte. Feedback obviously welcome!


src/main/java/kafdrop/util/Deserializers.java
```diff
@@ -0,0 +1,28 @@
+package kafdrop.util;
+
+public class Deserializers {
+
+    private MessageDeserializer keyDeserializer;
```
ekoutanov: Do we ever need to set these deserializers other than in the constructor? If not, can we make them `final`?

src/main/resources/templates/message-inspector.ftl
```diff
@@ -1,6 +1,6 @@
 <#ftl output_format="HTML">
 <#--
- Copyright 2016 Kafdrop contributors.
+ Copytright 2016 Kafdrop contributors.
```
ekoutanov: Typo.

src/main/java/kafdrop/service/KafkaHighLevelConsumer.java
```diff
@@ -138,12 +138,12 @@ private void initializeClient() {
   /**
    * Gets records from all partitions of a given topic.
    * @param count The maximum number of records getting back.
-   * @param deserializer Message deserializer
-   * @return A list of consumer records for a given topic.
+   * @param deserializers Key and Value deserializers
+   * @return A list of consumer records for a given tosic.
```
ekoutanov: Typo.

src/main/java/kafdrop/service/KafkaHighLevelConsumer.java
```diff
@@ -187,9 +187,8 @@ private void initializeClient() {
             rec.timestampType(),
             0L,
             rec.serializedKeySize(),
-            rec.serializedValueSize(),
-            rec.key(),
-            deserialize(deserializer, rec.value()),
+            rec.serializedValueSize(), deserialize(deserializers.getKeyDeserializer(), rec.key()),
```
ekoutanov: Would you mind having each arg on its own line, otherwise it gets hard to read with so many arguments.

ekoutanov (changes requested):

src/main/java/kafdrop/util/Deserializers.java
nicklester: Agree. Resolved in latest push

src/main/resources/templates/message-inspector.ftl
nicklester: Agree. Resolved in latest push

src/main/java/kafdrop/service/KafkaHighLevelConsumer.java
nicklester: Agree. Resolved in latest push

src/main/java/kafdrop/service/KafkaHighLevelConsumer.java
nicklester: Agree. Resolved in latest push

ekoutanov: Thanks @nicklester

Issue #86, opened 2020-02-20
Add possibility to remove a topic


stijnvanbever:
Hello,
It would be great to have the possibility to delete a topic.
While testing streaming applications I need to delete topics multiple times if my schema changes.
Would this fit into Kafdrop? Or is Kafdrop only meant as a read only tool?
Thanks

ekoutanov:
Not it all. While it started as a read-only tool, we had since added some write capabilities too — for example, to create topics. I think the community would appreciate having delete capability, if you're keen on raising a PR?

In the longer term, ideally Kafdrop should do most things that you currently need the CLI for.

stijnvanbever:
I'll see what I can do. PR will follow later

ekoutanov:
Any further thoughts on this @stijnvanbever ?

stijnvanbever:
Been a quite hectic time. I'm picking this up again now.

PR #108, opened 2020-04-17
Add possibility to remove a topic

stijnvanbever:
Added a button in the topic details that allows to delete a topic

Implements #86

### DIFF
--- a/src/main/java/kafdrop/service/KafkaHighLevelAdminClient.java
+++ b/src/main/java/kafdrop/service/KafkaHighLevelAdminClient.java
@@ -133,6 +133,25 @@ public final class KafkaHighLevelAdminClient {
     }
   }
 
+  /**
+   * Delete topic or throw ${@code KafkaAdminClientException}
+   *
+   * @param topic name of the topic to delete
+   * @throws KafkaAdminClientException if computation threw an Exception
+   */
+  void deleteTopic(String topic) {
+    DeleteTopicsOptions options = new DeleteTopicsOptions();
+    options.timeoutMs(5000); // timeout after 5 seconds
+    final var deleteTopicsResult = adminClient.deleteTopics(List.of(topic), options);
+    try {
+      deleteTopicsResult.all().get();
+      LOG.info("Topic {} successfully deleted", topic);
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.error("Error while deleting topic", e);
+      throw new KafkaAdminClientException(e);
+    }
+  }
+
   Collection<AclBinding> listAcls() {
     final Collection<AclBinding> aclsBindings;
     try {
--- a/src/main/java/kafdrop/controller/TopicController.java
+++ b/src/main/java/kafdrop/controller/TopicController.java
@@ -19,11 +19,8 @@
 package kafdrop.controller;
 
 import io.swagger.annotations.*;
-import kafdrop.config.KafkaConfiguration;
 import kafdrop.model.*;
 import kafdrop.service.*;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.http.*;
 import org.springframework.stereotype.*;
 import org.springframework.ui.*;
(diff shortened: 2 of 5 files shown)


ekoutanov:
There's an open PR that is earlier than this: https://github.com/obsidiandynamics/kafdrop/pull/97.

Is that essentially the same thing?

aaronwolmutt:
Please see #97 if you could.

src/main/java/kafdrop/service/KafkaHighLevelAdminClient.java
```diff
@@ -133,6 +133,25 @@ void createTopic(NewTopic newTopic) {
...
+   * @param topic name of the topic to delete
+   * @throws KafkaAdminClientException if computation threw an Exception
+   */
+  void deleteTopic(String topic) {
+    DeleteTopicsOptions options = new DeleteTopicsOptions();
+    options.timeoutMs(5000); // timeout after 5 seconds
+    final var deleteTopicsResult = adminClient.deleteTopics(List.of(topic), options);
+    try {
+      deleteTopicsResult.all().get();
+      LOG.info("Topic {} successfully deleted", topic);
```
ghost, line 148: Please see #97 . I believe #97 's implementation would log the topic name and topic partition number successfully deleted when the future is completed. It could also be overloaded to accept a collection of topic names.

stijnvanbever:
> There's an open PR that is earlier than this: #97.
> 
> Is that essentially the same thing?

There is a difference between the two.
In this PR the 'Delete topic' button is on the topic detail page.
In the other the button is on the overview.

I also see some problems with the other PR, if needed I can provide some feedback.
How would you like to continue?

src/main/java/kafdrop/service/KafkaHighLevelAdminClient.java
stijnvanbever: I'll check it out, thanks

stijnvanbever: having checked #97 I don't believe this is the case.
`deletionResult.values()` returns a map with key `String` and values `KafkaFuture<Void>`.
Your implementation would take the keyset (String) and print these values.
But with that implementation you're not checking if the futures have completed or not (they are asynchronous).
So it could be the case that either the Futures haven't completed yet or that they aren't successful and it would still log a successful deletion

ghost: Thanks, I will change #97 to invoke `KafkaFuture.get() ` so that `void deleteTopic(String topic) ` is async. My point was that logging the topic name and partition id is more effective than logging the string passed to `void deleteTopic(String topic) `

stijnvanbever: can you point me in the direction where #97 logs the partition id?
I only see it logging the topic name.
Also don't see the added value in highlighting that specific partitions are deleted.
Deleting a topic also deletes all of its partitions

ghost: I misread this, logging the ids of successfully deleted partitions would be useful if a timeout were to occur or for that ensuring replicated topics are deleted across a multi broker cluster. I think this is a good implentation: 

`private void tryDelete(AdminClient adminClient, String topic)
    throws Exception {
  try {
 adminClient.deleteTopics(Collections.singleton(topic)).all().get(DELETE_TIMEOUT_SECONDS, TimeUnit.SECONDS);
  } catch (TimeoutException e) {
    LOG.info("Did not receive delete topic response within %d seconds. Checking if it succeeded",
      DELETE_TIMEOUT_SECONDS);
    if (adminClient.listTopics().names().get(DELETE_TIMEOUT_SECONDS, TimeUnit.SECONDS).contains(topic)) {
      throw new Exception("Topic still exists after timeout");
    }
  }
}
`
See: 

https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaTestEnvironmentImpl.java#L168

PR #133, opened 2020-05-27
Some useful features

Wouter86:


ekoutanov:
@Wouter86 thanks for your effort.

Unfortunately I cannot accept this request due to its sheer size. There are four or five distinct features introduced and each should be assessed on its own merit. The workload for the maintainers to assess a PR of this size makes it unsustainable for us.

On the "view all partitions" feature, I don't understand how this works conceptually, as the concept of 'start offset' does not make sense for a topic. How does one navigate through the messages?

Can I suggest that future PRs are broken down by distinct areas of functionality.

For highly involved features such as viewing all messages, it it customary to discuss the design with the project's maintainers prior to commencing work. This avoids surprises and wasted time in both implementation and review.

Wouter86:
Hi Emil,

Sorry to hear you cannot accept the PR. Basically, I needed the features fast, so I just added them. On the other hand, I also wanted to share them.

It is, however, the first time I do a PR on github, so excuse me for the size.

Viewing messages from any partition allows to fetch the latest messages for the topic. In this case, the start offset indeed does not make sense and is ignored. By selecting "any" partition and specifying a message count, you basically fetch the latest messages of the topic, limited by the given count per partition (<=99999/partition). Results are returned in descending timestamp order for convenient browsing.

Other then that, the implemented features are:
- Message Filtering support: 'contains' on key, message, header
- Configuration support to hide the Delete Topic Button
- Configuration support to hide elaborate topic information shown by default throughout the app and as such, increase UI performance.

I really needed the last feature to use kafdrop on large clusters with many topics, partitions and consumers, because the performance of application navigation was otherwise unacceptably slow. It is a "quick" implementation though. A thorough implementation may, e.g., restructure the application UI or add caching to avoid the performance issue.

Are you interested in any of the features?

ekoutanov:
The delete button has been implemented by one of the maintainers in a recent PR #135 

I would say that filtering would be somewhat useful, but in its present state it takes up too much screen real estate. Perhaps if the filter controls were concealed in a roll-up of some sort.

Wouter86:
I'm no UI expert so would it be possible for the UI part to be enhanced as required by someone else? For me and my user group the current filter implementation is already a decent start.

I really hope the other features will find their way into the product as well, one way or another. For us, they are really crucial for the adoption of kafdrop.

Wouter86:
@ekoutanov ?

davidetan:
I have to admit that 

> Configuration support to hide elaborate topic information shown by default throughout the app and as such, increase UI performance.

is necessary for me and for large clusters users, otherwise kafdrop is not usable.
Everytime I go to the homepage it takes more than one minute to load all topic information.
This is unfortunate since this dashboard is exactly what I need. Still is not usable for me :(

Wouter86:
@ekoutanov can you give an update?

As stated earlier, I'm no UI expert so would it be possible for the UI part to be enhanced as required by someone else? For me and my user group the current filter implementation is already a decent start. 

Do I make a branch featuring it for a PR?

PR #142, opened 2020-06-04
Reduce number of calls to Kafka

jchipmunk:
Kafdrop uses a heavy topic info object in all cases, even if additional data is not needed on UI side. It is proposed to add the ability to manage enrichment of the topic info object using the following modes:
* `PartitionSize` - specifies to enrich topic info with partition size (first offset, last offset and size)
* `TopicConfig` - specifies to enrich topic info with topic configuration

These changes are based on the following assumptions:
* `cluster-overview` and `broker-detail` views do not need data about size of partitions
* `message-inspector` and `consumer-detail` views do not need data about topic configuration

### DIFF
--- a/src/main/java/kafdrop/controller/BrokerController.java
+++ b/src/main/java/kafdrop/controller/BrokerController.java
@@ -40,7 +40,7 @@ public final class BrokerController {
   public String brokerDetails(@PathVariable("id") int brokerId, Model model) {
     model.addAttribute("broker", kafkaMonitor.getBroker(brokerId)
         .orElseThrow(() -> new BrokerNotFoundException("No such broker " + brokerId)));
-    model.addAttribute("topics", kafkaMonitor.getTopics());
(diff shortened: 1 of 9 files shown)


xakassi:
Please, merge this. It greatly improved the speed of page loading!

ekoutanov:
I like the idea of only pulling what we need from Kafka; however, I have some reservations regarding the implementation.

The code in the PR is trying to reuse the same data models, and is performing "enrichment" on them, which makes it hard to follow, and also results in a bunch of fields that are uninitialised. I would much rather us split up the value objects and use "lighter" VOs where appropriate, without any enricher methods.

jonas-grgt:
I must agree with @ekoutanov. Also adding a kind of flag argument to alter the behaviour of a method isn't a nice solution. I'm refering the clean code here. 
Next to that I was looking at the code and noticed there was actually no need for the deviation since the `configMap` of the `TopicVo` is never used. Or am I overlooking something here ? If not I believe removing the code that fills up the unused `configMap` would speed up the page loading also and would be less invasive.

PR #145, opened 2020-06-08
Read Message from all partitions

ctrongduc:
Add option to view message from ALL partitions


ekoutanov (commented):
Haven't ran the change yet, but looking at the implementation I don't think this is going to work. When dealing with multiple partitions in a topic, the concept of a single offset does not really make sense. There are a number of offsets — one for each partition, and the navigation from one batch of messages to the next needs to take this into account.

ctrongduc:
Could you help me understand the feature a little more? 
The way I see it is, given an offset and number of message to poll, the program looks for messages in ALL partitons from the given offset up to number of message, this is very similar to reading messages from one partition.
My solution to this is to get number of message from All partitions then order by offset, get top n requested messages. This is not the best way but it does the job.
I might have misunderstood something here, please enlighten me. Thanks.

I'm planning to have this fixed over the weekend, if thing get cleared.

ekoutanov:
The main issue is that there are multiple partitions, each with its own set of offsets. Furthermore, the partitions might not be evenly balanced, in that some partitions might receive a lot of messages, others might be lightly loaded.

As an example, consider a topic with two partitions p0 and p1:
p0 is heavily loaded, with messages occupying offsets 300000 to 400000.
p1 is lightly loaded, with messages in the offset range 0 to 100.

Not only partitions are completely unrelated, their offset ranges don't even have to overlap.

How does the user navigate these partitions? There really is no concept of a starting offset for a topic.

The way I see this solved is to provide the user with a *completely different view* for navigating topics vs partitions. In other words, we don't attempt to touch the existing partition view — we create an entire new screen for browsing a topic, with an entirely new set of models, controllers, APIs, etc.

Whereas a partition viewer displays N records from some starting offset, a topic viewer works in reverse. It starts from the end of the topic, and displays the N most recent messages across all partitions in a merged view. (Sorted in reverse chronological order.). At the bottom of a screen there might be a "More records" button, which fetches additional records and displays them.

An alternative approach is to use time, rather than offsets. In other words, the user starts from a point in time, and we use Kafka's API to translate a timestamp into a vector of offsets, the read from all partitions starting from that offset.
 
At any rate, I don't think building on the current PR will cut the mustard. This is an entirely new feature that should not touch any of the existing ones.

ctrongduc:
Ah, I got the idea. At first I thought the feature should not deviate too much from current view. Your description is much more than that. With this new view, we actually could do the constant refresh on a topic, which what I really want to do. 
Lets close this PR and I'll start over.

ekoutanov:
Yeah I think it's a good opportunity to bring in completely new capabilities to Kafdrop.

siripurapugopi:
Any plans on adding this feature?

PR #146, opened 2020-06-08
Updated the paths for .js and .css to not be absolute paths 

FrozenInc:
This makes it possible for the paths to be found behind a reverse proxy. As right now if a webserver takes care of the url paths and puts kafdrop some place else than /, kafdrop will not be able to find the .css and .js files as it uses absolute paths /js and /css.
By switching over to ./js and ./css, kafdrop will be able to handle paths behind a reverse proxy.

### DIFF
--- a/src/main/resources/templates/acl-overview.ftl
+++ b/src/main/resources/templates/acl-overview.ftl
@@ -27,7 +27,7 @@
     </style>
 </@template.header>
 
-<script src="<@spring.url '/js/powerFilter.js'/>"></script>
+<script src="<@spring.url './js/powerFilter.js'/>"></script>
 
 <#setting number_format="0">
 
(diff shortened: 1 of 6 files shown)


pankti11:
Hey @FrozenInc ,
This does not seem to work. kafdrop works behind the proxy only with root url not with the subpath.

ingress configuration:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: kafdrop-ui
  namespace: ix
  annotations:
   ingress.kubernetes.io/rewrite-target: /
spec:
  rules:
    - http:
        paths:
        - backend:
            serviceName: kafdropui
            servicePort: 80
          path: /kafdrop/
```

image

Any help with the configuration will be a great help. Thanks you !

PR #148, opened 2020-06-11
Configure topic creation with --topic.createEnabled

ohjongsung:
Don't show the New button when --topic.createEnabled=false
The button is visible by default, no --topic.createEnabled is required

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,12 @@ By default, you could delete a topic. If you don't want this feature, you could 
 --topic.deleteEnabled=false
 ```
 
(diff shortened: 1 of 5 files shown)


ekoutanov (changes requested):
We need to make sure that the create-enabled/disabled check is performed not only on the UI, but also in the controller code. I.e. the user shouldn't be allowed to bypass the UI and create the topic, if this has been disabled.

ohjongsung:
--topic.createEnabled flag is also checked in the controller code.

Issue #167, opened 2020-07-21
Update kafka-avro-serializer dependency


typhoon2k:
Hi,

while testing kafdrop with the latest version of Kafka stack I've discovered that kafdrop is not able to display messages with references in schemas (more details about this functionality can be found in [here](https://docs.confluent.io/current/schema-registry/serdes-develop/index.html#referenced-schemas)). Solution to this problem is to update kafka-avro-serializer dependency to the latest version (at the moment of writing this latest version is 5.5.1).
Please release a version of kafdrop with upgraded kafka-avro-serializer.

ty

typhoon2k:
@ekoutanov , please review/approve PR.

typhoon2k:
Already fixed in `master`.

PR #185, opened 2020-09-11
Update version of kafka-avro-serializer (#167)

typhoon2k:
Update version of kafka-avro-serializer to support schema references as described in https://docs.confluent.io/current/schema-registry/serdes-develop/index.html#referenced-schemas (https://github.com/obsidiandynamics/kafdrop/issues/167).


typhoon2k:
`master` already contains required changes.

Issue #201, opened 2020-10-25
Kafdrop connection to an kafka servier with just SSL (not SASL) does not work


krishna-prasad-s:
In the readme its described as its working, but when I provide the required details. It throws an exception (mainly because the server refuses)
the docker command
```
docker run  --rm -p 9000:9000 \
    --name kafdrop \
    -e KAFKA_BROKERCONNECT="172.17.0.3:9092" \
    -e KAFKA_ISSECURED="false" \
    -e KAFKA_PROPERTIES="$(cat kafka.properties | base64)" \
    -e KAFKA_TRUSTSTORE="$(cat ../truststore/kafka.truststore.jks | base64)" \
    -e KAFKA_KEYSTORE="$(cat ../kafdrop-keystore/kafdrop.keystore.jks | base64)" \
    -e JVM_OPTS="-Xms32M -Xmx64M" \
    -e SERVER_SERVLET_CONTEXTPATH="/" \
    obsidiandynamics/kafdrop
```

```
my kafka.properties
ssl.security.protocol=SSL
ssl.endpoint.identification.algorithm=
ssl.protocol=TLS

ssl.truststore.password="*****"
ssl.truststore.type=JKS

ssl.keystore.password="*****"
ssl.keystore.type=JKS
```
the exception

```
2020-10-25 14:26:57.533  INFO ${sys:PID} [           main] k.Kafdrop$EnvironmentSetupListener       : Initializing JAAS config
2020-10-25 14:26:57.542  INFO ${sys:PID} [           main] k.Kafdrop$EnvironmentSetupListener       : env: null .isSecured kafka: false
2020-10-25 14:26:57.543  INFO ${sys:PID} [           main] k.Kafdrop$EnvironmentSetupListener       : Env: null
2020-10-25 14:26:57.681  INFO 1 [           main] o.s.b.StartupInfoLogger                  : Starting application on 7d0eb263b4cc with PID 1 (started by root in /)
2020-10-25 14:26:57.682  INFO 1 [           main] o.s.b.SpringApplication                  : No active profile set, falling back to default profiles: default
2020-10-25 14:26:59.035  INFO 1 [           main] i.u.s.s.ServletContextImpl               : Initializing Spring embedded WebApplicationContext
2020-10-25 14:26:59.036  INFO 1 [           main] w.s.c.ServletWebServerApplicationContext : Root WebApplicationContext: initialization completed in 1314 ms
[...]

PR #202, opened 2020-10-26
move security protocol out of isSecured block

krishna-prasad-s:
this resolves the issue that I mentioned in
https://github.com/obsidiandynamics/kafdrop/issues/201
 I have tested with SSL, with SASL_SSL and without SSL it works in all three cases.


src/main/java/kafdrop/config/KafkaConfiguration.java
```diff
@@ -28,10 +28,13 @@ public void applyCommon(Properties properties) {
     properties.setProperty(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, brokerConnect);
     if (isSecured) {
       LOG.warn("The 'isSecured' property is deprecated; consult README.md on the preferred way to configure security");
-      properties.put(CommonClientConfigs.SECURITY_PROTOCOL_CONFIG, securityProtocol);
       properties.put(SaslConfigs.SASL_MECHANISM, saslMechanism);
     }
 
+    if(isSecured || securityProtocol.equals("SSL")) {
```
davideicardi, line 34: I don't understand why this is required only for SSL.
From [doc](https://kafka.apache.org/24/javadoc/org/apache/kafka/common/security/auth/SecurityProtocol.html) security protocol can be:
- PLAINTEXT
- SASL_PLAINTEXT
- SASL_SSL
- SSL

Why here we just check SSL?

davideicardi: cc @Bert-R

krishna-prasad-s: we are checking for isSecured or SSL and not just for SSL.
isSecured already covers for SASL. what remains is that we apply the security configuration in case of just SSL. This is what is done.  otherwise mTLS was only working for SSL_SASL and not just SSL.

we don't need security configuration for PLAINTEXT.

davideicardi: One more question: Why `isSecured` is false in case of `SSL`?

P.S. Sorry to ask but I'm not an expert, just trying to maintain the project ...

Bert-R: The ``isSecured`` property is old. It's deprecated since 2019, see line 30. So that should generally not be set to ``true``.

davideicardi: Ok, so it is probably better to only check `securityProtocol` and ignore `isSecured`?

Bert-R: That would break the legacy `isSecured` feature, as the property put is removed from line 31.

krishna-prasad-s: isSecured is pulled from the application config, look at kafdrop.java:127 and application.yaml:35. so in case someone does set it true. It needs to work, which is why I kept it as is.  We can raise a separate issue to remove the whole isSecured part and then clean it up.

davideicardi (approved):

PR #211, opened 2020-11-13
Update to Spring Boot 2.2.x

janderssonse:
This change does an basic update to Spring Boot 2.2.10 and booterizes some dependencies to be handled under Spring Boot's dependency managment. This will make future updates easier, as the components in a Spring Boot release cycle is (somewhat) tested to be compatible. The reason this update did not go to Spring Boot 2.3.x at once is that I think it isbetter to take small steps to iron out upgrade bugs.


kchibui (approved):
LG

PR #242, opened 2021-02-18
avoid deleting configuration files

sopak:
In case $KAFKA_PROPERTIES $KAFKA_TRUSTSTORE or $KAFKA_KEYSTORE exists,
they will overwrite its configuration file.
Deleting them if they are not present is preventing to prepare them directly.
For example by init container in kubernetes where I do need to pass passphrase from secret created by cert-manager.


ChristianCiach:
Please merge this! I've no idea why kafdrop forces us to pass these things as base64-encoded environment variables. Why can't we just mount `kafka.properties` and the other files directly into the container?!

Bert-R:
This is addressed now through #502. Sorry for not noticing your PR earlier. I only saw it while working on #502.

PR #259, opened 2021-03-29
updated dockerfile

jhonDoe15:
updated Dockerfile to a newer more secure base image


davideicardi:
We now use `eclipse-temurin:11` as base image.

PR #273, opened 2021-04-23
 add terraform "Running in Kubernetes (using Terraform and Helm Chart) "

reza-gholizade:
the best way to install Kafdrop on Kubernetes


OneCricketeer:
Why not use the Helm CLI to install? What's the purpose of Terraform here beyond what Helm already offers?

PR #278, opened 2021-05-17
add keyFormat and format query params to the view message link for topics

### LINKED ISSUE #239: Messages not read in AVRO format when navigating directly to a partition
When clicking directly on a partition go view the messages these messages aren't read in the Format that is selected.

image

As you can see the messages aren't being decoded
image

While the drop down says it selected AVRO
image

The url that go do to is something like `http://kafkadrop/topic/{{ topicname }}/messages?partition=6&offset=0&count=100`

It turns our that the `--message.format=AVRO` setting is not added to the partition url since its missing the `&keyFormat=AVRO&format=AVRO` params

jorkzijlstra:
I noticed that the view message for topic did not have the keyFormat and format query params specified which causes the message to be de-serialized in the "DEFAULT" format instead of the message.format specified in the config.

Fix #239

### DIFF
--- a/src/main/resources/templates/topic-detail.ftlh
+++ b/src/main/resources/templates/topic-detail.ftlh
@@ -124,7 +124,7 @@
                 <tbody>
                 <#list topic.partitions as p>
                     <tr>
-                        <td><a href="<@spring.url '/topic/${topic.name}/messages?partition=${p.id}&offset=${p.firstOffset}&count=100'/>">${p.id}</a></td>
+                        <td><a href="<@spring.url '/topic/${topic.name}/messages?partition=${p.id}&offset=${p.firstOffset}&count=100&keyFormat${keyFormat}=&format=${format}'/>">${p.id}</a></td>
                         <td>${p.firstOffset}</td>
                         <td>${p.size}</td>
                         <td>${p.size - p.firstOffset}</td>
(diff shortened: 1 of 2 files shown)


davideicardi (approved):
LGTM! Thank you!

Do you think it is possible to write a simple unit test to verify this bug?

jorkzijlstra:
Not sure what / how to test this. It's controller logic and I don't see any other tests for them. Also not familiar with the test framework yet so I'll have to have a look later.

davideicardi:
@jorkzijlstra ok for now I think it should be fine.

Can you just confirm that this change doesn't have any effect if `--message.format=AVRO` is not used? It works as before without problems?

jorkzijlstra:
@davideicardi I can confirm that since that is also what we are actually running. It will just revert to the `DEFAULT`

image

src/main/resources/templates/topic-detail.ftlh
```diff
@@ -124,7 +124,7 @@
                 <tbody>
                 <#list topic.partitions as p>
                     <tr>
-                        <td><a href="<@spring.url '/topic/${topic.name}/messages?partition=${p.id}&offset=${p.firstOffset}&count=100'/>">${p.id}</a></td>
+                        <td><a href="<@spring.url '/topic/${topic.name}/messages?partition=${p.id}&offset=${p.firstOffset}&count=100&keyFormat${keyFormat}=&format=${format}'/>">${p.id}</a></td>
```
fape, line 127: I think that sould be `keyFormat=${keyFormat}&format=${format}`

PR #291, opened 2021-07-22
Use configurable $SERVER_PORT for management endpoints

odin-delrio:
If using the `$SERVER_PORT` environment variable for configuring the server port, management endpoints still use always the port **9000**.

I think it would be nice to also use the `$SERVER_PORT` value there.

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -21,7 +21,7 @@ management:
       base-path: /actuator
       exposure.include: "*"
   server:
-    port: 9000
+    port: ${SERVER_PORT:9000}
 
 kafdrop.monitor:
   clientId: Kafdrop


src/main/resources/application.yml
```diff
@@ -21,7 +21,7 @@ management:
       base-path: /actuator
       exposure.include: "*"
   server:
-    port: 9000
+    port: ${SERVER_PORT:9000}
```
odin-delrio, line 24: doing the same as in the line 2 in this file.

daniel-dios (approved):

src/main/resources/application.yml
eydunn: Then it rather should be:   ` port: ${MANAGEMENT_SERVER_PORT:9000}`, but actually both are redundant. They should only be
 
```
server
    port: 9000
    
management
    server
        port: 9000

```
Because the env variables SERVER_PORT and MANAGEMENT_SERVER_PORT will override these properties anyway.

Bert-R: Thanks for mentioning it. Closing this PR. For future reference: see [this](https://docs.spring.io/spring-boot/docs/1.5.6.RELEASE/reference/html/boot-features-external-config.html#:~:text=you%20can%20use%20underscores).

PR #314, opened 2021-12-08
Move to latest confluent packages to have latest protobuf functionaity ty in schema registry

venkatduddu:
Confluent released fix for referencing the protobuf schemas as part of 6.2. Without this functionaliy I am seeing below error in Kafdrop UI

```
org.apache.kafka.common.errors.SerializationException: Error deserializing Protobuf message for id 65 Caused by: java.lang.IllegalStateException: com.google.protobuf.Descriptors$DescriptorValidationException: rio.event.WebRequest.req_timestamp: "google.protobuf.Timestamp" is not defined. at
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,7 +89,7 @@
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>kafka-avro-serializer</artifactId>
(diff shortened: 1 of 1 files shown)


davideicardi:
See #321

Issue #323, opened 2021-12-23
Restore CI/CD


davideicardi:
It looks like Travis CI doesn't work anymore. How we can restore it?
Can we eval to use github actions instead?

It is important to also release the new docker image with Log4J vulnerability fix.

@ekoutanov  do you have any suggestion?

cemeyer2:
for now, I wrote my own dockerfile and pushed it to a private ECR to unblock. Not the most elegant but works for now until an official image is pushed

```Dockerfile
FROM openjdk:11 AS builder

RUN mkdir /app && cd /app && \
    wget https://github.com/obsidiandynamics/kafdrop/archive/refs/heads/master.zip && \
    unzip master.zip && \
    cd /app/kafdrop-master && \
    ./mvnw -DskipTests package assembly:single && \
    mkdir /kafdrop && \
    cp /app/kafdrop-master/target/kafdrop*tar.gz /kafdrop.tar.gz && \
    cp /app/kafdrop-master/target/docker-ready/kafdrop.sh / && \
    chmod +x /kafdrop.sh

FROM openjdk:11
COPY --from=builder /kafdrop.tar.gz /
COPY --from=builder /kafdrop.sh /
RUN tar xvf /kafdrop.tar.gz
CMD ["/kafdrop.sh"]
```

davideicardi:
Thank you @cemeyer2 , I have done something similar also for myself. But I have used `mvn assembly:single docker:build`, that, as described [here](https://github.com/obsidiandynamics/kafdrop), can create the docker image using the official workflow.
In the following days I will try to investigate how to fix CI/CD and hope to fix it soon...

ecetiner87:
Thank you @cemeyer2 and @davideicardi for this help. We built the image from related dockerfile that @cemeyer2  shared and it works properly. Security scan is also passed. Much appreciated.

niqbal:
@davideicardi i tried your approach , however I am noticing that i still get this `spring-boot-starter-log4j2-2.6.1.jar` which is marked as vulnerable in the maven repository: 
https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-log4j2/2.6.1

The log4j jars have been upgraded, so i dont think it should be an issue, but wanted to confirm.

````
root@kafdrop-596d547c7-tgmb4:/# find .  | grep jar | grep log4j  | xargs ls -al 
[...]

Issue #260, opened 2021-04-01
Bintray going away may 1st 2021


jgrammen-agilitypr:
Bintray service is being shutdown may 1st 2021
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

Kafdrop uses bintray to distribute the .deb binaries, kafdrop will need to migrate to new method of provided deb compatible repo's.

jgrammen-agilitypr:
This is now critical as the sunset will happen in 3 days

jgrammen-agilitypr:
You can now no longer download kafdrop as bintray is offline

jgrammen-agilitypr:
emailed user "ekoutanov" who appears to be the primary contact for the repo, no response as of yet.
This currently no way to download pre compiled versions of this repository.

Alfusainey:
In  the meantime,  the alternatives for running kafdrop locally are (1) [using docker](https://github.com/obsidiandynamics/kafdrop#running-with-docker) or (2) clone & build the source with maven (see [building section](https://github.com/obsidiandynamics/kafdrop#building) of the readme)

smshr:
The build using command `mvn clean package` fails with error: 
```
Caused by: org.apache.maven.plugin.compiler.CompilationFailureException: Compilation failure
    at org.apache.maven.plugin.compiler.AbstractCompilerMojo.execute (AbstractCompilerMojo.java:1224)
    at org.apache.maven.plugin.compiler.CompilerMojo.execute (CompilerMojo.java:187)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:208)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
```
I am using OpenJDK16.

combrs:
Hello, no solution for now? Wants to download jar release also, and can't
Looks like abandoned project, only source code accessible
No pull requests was accepted in current year

pas8sion:
> The build using command `mvn clean package` fails with error:
> 
> ```
> Caused by: org.apache.maven.plugin.compiler.CompilationFailureException: Compilation failure
>     at org.apache.maven.plugin.compiler.AbstractCompilerMojo.execute (AbstractCompilerMojo.java:1224)
[...]

PR #325, opened 2021-12-28
added github actions to build and release

kbudde:
Hi,

I've added a github action to build and deploy as travis was doing it before.
You must add the following secrets to your repository:
- DOCKERHUB_USERNAME
- DOCKERHUB_TOKEN

#323


davideicardi (approved):
LGTM! Thank you very much!

Maybe we should delete `.github/workflows/maven.yml` ?

I don't have permission to setup github actions variables, I think we need @ekoutanov  .

kbudde:
The maven.yml will also run on PRs. Therefore I think it's better to keep it.

mcs:
Feel free to do whatever you want with the maven.yml. I just added it to my former PR because there "just" was Travis CI which I could not use for server-side test execution.

.github/workflows/build.yml
```diff
@@ -0,0 +1,90 @@
...
+      name: Login to Docker Hub
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }} 
+    -
+      name: push snapshot version
+      run: |
+        echo publishing docker container $app_ver
+        docker push docker.io/obsidiandynamics/kafdrop:$app_ver
```
mcs: Couldn't this accidentally overwrite an existing Dockerhub version if one forgets to add "-SNAPSHOT" after a release commit? Afaik this push with a new image id would succeed, while the Github release artifact wouln't change (see line 84).

Maybe it's better to first upload the Github release artifact and only if that succeeds, upload to Dockerhub and Bintray?

kbudde: I like the idea. @mcs  I updated the PR to the order: build, github release, bintray and docker hub.

kbudde:
@ekoutanov I would really love to see a release without log4j issues. 
Can you please set up the secrets?

ekoutanov:
I'll try to get around to it.

davideicardi:
@kbudde @ekoutanov @mcs 
According to #260  Bintray is no longer available. Maybe we should use Github release only?

https://github.com/marketplace/actions/upload-files-to-a-github-release

ekoutanov:
> @kbudde @ekoutanov @mcs According to #260 Bintray is no longer available. Maybe we should use Github release only?
> 
> https://github.com/marketplace/actions/upload-files-to-a-github-release

yeah, Bintray's kaput. We can skip it if the binaries can be uploaded elsewhere.

kbudde:
@davideicardi , @ekoutanov I fixed it and updated the description.

The github workflow will now create the github release and a tag(!) if the version in pom.xml does end on "-SNAPSHOT". 
The github release has the kafdrop-*VERSION*.jar attached.
See here for an example: https://github.com/kbudde/kafdrop/releases/tag/3.28.0

@ekoutanov only the dockerhub token is needed anymore. 
Bintray was removed. For github the default token is used.

I've also cleaned up the CI files.

Release workflow:
- Update pom.xml (remove `-SNAPSHOT` from version)
- Commit and push to master
- Update pom.xml (increment version + add suffix `-SNAPSHOT` again)
- Commit and push

Each commit on master will push container to dockerhub: `obsidiandynamics/kafdrop:*version*` (e.g. `obsidiandynamics/kafdrop:3.28.0-SNAPSHOT`, `obsidiandynamics/kafdrop:3.28.0`)
If the version is not a snapshot release, another image is pushed:  `obsidiandynamics/kafdrop:latest`.
So latest will always point to last release.

davideicardi:
Great work @kbudde ! Thank you!

ekoutanov:
Added secrets

davideicardi:
@kbudde The new action seems to work without problems!
I just released a new version 3.28.0.

PR #333, opened 2022-01-10
remove unsafe cast and redundant @ExtendWith

vojkog:
Constructor of `MapPropertySource` receives ` Map<String,Object> source` so we may avoid unsafe casting: 

Javadoc: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/env/MapPropertySource.html#MapPropertySource-java.lang.String-java.util.Map-

Also:  `@ExtendWith(SpringExtension.class)` is already included in `@SpringBootTest`:

Javadoc: https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/test/context/SpringBootTest.html


davideicardi (approved):
LGTM

PR #339, opened 2022-01-19
Override log4j to 2.17.1 to address CVE-2021-44832

Leszek-Kowalski:
Log4j v2.17.0 is vulnerable to CVE-2021-44832 and as such it`s being flagged in conservative organisations.
Upgrading to 2.17.1 to resolve and be able to continue using Kafdrop.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,6 +23,7 @@
         <protobuf.version>3.19.1</protobuf.version>
         <testcontainers.version>1.16.2</testcontainers.version>
         <kafka-libs.version>6.1.4</kafka-libs.version>
+        <log4j2.version>2.17.1</log4j2.version>
     </properties>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -119,6 +120,11 @@
             <artifactId>msgpack-core</artifactId>
             <version>0.8.24</version>
         </dependency>
+        <dependency>
```
davideicardi: Do we need this dependency? According to https://spring.io/blog/2021/12/10/log4j2-vulnerability-and-spring-boot?s=09 I think we need to only set the above `log4j2.version` property.

Leszek-Kowalski: You might be right. I am not good enough on spring magic...
Corrected

PR #348, opened 2022-02-08
release 3.30.0

jorkzijlstra:
Could the current version released to docker?

Not sure what the release step are, but noticed in the github workflow that at least this version should be updated. Probably after this merge it should be set to the next SNAPSHOT version again


jorkzijlstra:
Wrong version tag

PR #350, opened 2022-02-08
Prepare release 3.29.0

jorkzijlstra:
Could the current version released to docker?

Probably after this merge it should be set to the next SNAPSHOT version again


davideicardi (approved):
LGTM

For reference about release workflow see this: https://github.com/obsidiandynamics/kafdrop#release-workflow

PR #359, opened 2022-03-10
Add a favicon

brsolomon-deloitte:


davideicardi:
Thank you @brsolomon-deloitte !

It the favicon automatically used? Or do we need some additional work to use it?

brsolomon-deloitte:
@davideicardi I thought it would work out of the box because it is on the static resources path, but it appears it does not, at least when serving through a locally built docker image. :( Going to mark as draft until I can figure out what else is needed.

PR #382, opened 2022-05-31
Support Protobuf "Any" message type for unmixed topics

gebinic:
- add select box for "Any" message type support
- all messages must be either "Any" or not "Any" in a topic


davideicardi (changes requested):
Thank you for the contribution!

Can you please just explain a little bit the changes you have made and which problems this will solve?
I never used protobuf.

gebinic:
It is a common practice to put messages from one specific type (protos in this case) in a topic but sometimes it might be better to put messages from several types in a topic, depending on your requirements. From a consumer's perspective it is not possible to handle multiple message types in one topic. Protobuf's "[Any](https://developers.google.com/protocol-buffers/docs/proto3#any)" message type allows developers to wrap each proto and send its bytes to the topic. Consumers can treat these bytes as one specific type, i.e as "Any" message type and unwrap the real message from it.

The problem with the currently provided ProtobufMessageDeserializer implementation is that it simply parses each incoming byte buffer as [DynamicMessage](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/DynamicMessage) and prints it as a json string, but takes no care about the "Any" message type. The current default string representation for "Any" is only its containing message type url, but not the real message.

So, my solution in a nutshell:
- select "Yes" in the "Is Any proto?" select box if the topic contains "Any" message types
- ProtobufMessageDeserializer gets a new flag at runtime which says whether the incoming byte buffer will be an "Any" message type or not
- if this flag is true, the byte buffer is treated as "Any" message type and its containing value, i.e the real kafka message is printed as json string instead of the "Any" message type itself
- if the flag is false (so, not an "Any" message type), the current behavior takes place

davideicardi (approved):
LGTM!
@ekoutanov  Do you have any comment on this?

PR #383, opened 2022-05-31
Pre-select a descriptor file equal to topic name if available

gebinic:


davideicardi (changes requested):
Thank you for the contribution!
Can you please describe a little bit the feature?

The assumption is that there is a descriptor file with the same name of the topic?

gebinic:
> The assumption is that there is a descriptor file with the same name of the topic?

Yes! This is a known limitation of this convenience feature. If such a descriptor file could not be found, no one will be pre-selected in kafdrop's web ui, i.e the currently known behavior takes place.

Lets assume a kafka topic name is "example" and there is a list of descriptor files "example.desc" and "example2.desc". When a developer requests the kafka messages, the corresponding method in the controller class compares each descriptor file name with the given topic name. If there is a match, this descriptor file will be pre-selected in kafdrop's web ui. Otherwise, an arbitrary descriptor file is pre-select (default behavior).

davideicardi (approved):

PR #385, opened 2022-06-13
Add configurable support for Kubernetes 1.22 Ingress v1beta1 obsoletion

andytson:
Note `helm template` by default will have `.Capabilities.KubeVersion` based on a hard-coded test value without `--kube-version` set. helm install/upgrade will detect from the cluster.

Alternatively `global.kubeVersion` or `ingress.apiVersion` be set


davideicardi (approved):
Thank you!

PR #386, opened 2022-06-14
Add configurable pathType and ingressClassName to Ingress

andytson:
The previous PR I missed completing the pathType changes I was meaning to add support for.

Now we have ingressClassName as well, which is much more relevant when using v1 Ingresses (e.g. ingress-nginx requires use of IngressClass for them)

pathType was optional in v1beta, and not available before 1.18, but is mandatory in v1, however changing to the previous default pathType that would be used for backwards-compatibility

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -32,7 +32,7 @@ spec:
                 port:
                   name: http
           {{- else }}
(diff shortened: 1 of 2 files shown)


davideicardi (approved):
Thank you!

PR #391, opened 2022-06-17
Disable HTTP TRACE and TRACK methods

tbowman-sfmc:
Corporate Security Scanners find both the HTTP TRACE and TRACK methods to be exploitable so this modifies the default behavior of Undertow to disallow those methods.  I've included a test to ensure that rejection occurs.

### DIFF
--- a/src/main/java/kafdrop/controller/ClusterController.java
+++ b/src/main/java/kafdrop/controller/ClusterController.java
@@ -121,5 +121,16 @@ public final class ClusterController {
     ClusterSummaryVO summary;
     List<BrokerVO> brokers;
     List<TopicVO> topics;
+    public ClusterSummaryVO getSummary() {
+      return summary;
+    }
+
+    public List<BrokerVO> getBrokers() {
+      return brokers;
+    }
+
+    public List<TopicVO> getTopics() {
+      return topics;
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/kafdrop/controller/ClusterController.java
```diff
@@ -121,5 +121,16 @@ public static final class ClusterInfoVO {
...
+    public ClusterSummaryVO getSummary() {
+      return summary;
+    }
+
+    public List<BrokerVO> getBrokers() {
+      return brokers;
+    }
+
+    public List<TopicVO> getTopics() {
+      return topics;
```
davideicardi, line 133: Why do we need these new functions?

davideicardi (commented):
Thank you!  LGTM! Just one question:
There is a specific security issues regarding Trace and Track methods? Or is just a best practice?

tbowman-sfmc:
Trace has been used to steal user credentials so the recommendation is to disable it by default: https://owasp.org/www-community/attacks/Cross_Site_Tracing

Track is IIS specific but it gets lumped together with Trace since it shares the same vulnerabilities.

tbowman-sfmc:
Sorry, I missed your question at the bottom about the functions. The VO object would error out unless those functions were defined so I added them to get a successful test run.

PR #402, opened 2022-07-30
Uses fabric8io docker maven plugin

altaiezior:
The purpose of this MR is to move to fabric8io docker maven plugin as spotify docker maven plugin is no longer maintained and doesn't provide arm64 compatibility.

With the given setup, one can change the base image using property -Ddocker.base.image while building.

`eclipse-temurin` doesn't have arm64 compatible manifest as of now but has arm images under `arm64v8/eclipse-temurin`.

Also github workflows have runners based on amd64 processors while circleci has runners available for both arm64 and amd64 machines.

So in all it would be a tough setup to provide arm compatible images but this MR would still help someone who is building locally.

On the other hand in my fork I can work on to provide an arm64 compatible counterpart of kafdrop in the interim a solution is being discovered.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -26,7 +26,7 @@ jobs:
         echo "app_ver=$(mvn -B help:evaluate -Dexpression=project.version -q -DforceStdout)" >> $GITHUB_ENV
     -
       name: Build with Maven
-      run: mvn -B clean integration-test package assembly:single docker:build
+      run: mvn -B clean integration-test package docker:build
 
 # github_release:
     -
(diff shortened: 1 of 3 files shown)


nelopuchades:
Hi @anuragagarwal561994!

I'm trying to build-and-run this image in a docker compose locally together with the kafka image from obsidian. While they review this MR (i'm not very optimistic with this 😞), how are you being able to do it "locally" in an ARM-machine? My docker-compose doesn't work if this image is there 🤔 

I'm trying to do it in an Apple Macbook Pro M2.

Thank you very much!

altaiezior:
Hey @nelodev can you help me with a screenshot of the issue and the error you are getting.

I think I know the issue. This change itself doesn't itself make the image arm compatible I believe.

This is just the first step. I have a fork and the release page on the same where I have published the image.

https://github.com/arm64-compat/kafdrop

You can try that and let me know if it works.

I don't think that M2 should create a problem, yet if you can provide me the error I can try looking into it.

Issue #403, opened 2022-08-10
CVEs in the 3.30.0


cbl315:
I scan the image `obsidiandynamics/kafdrop:3.30.0`, and find many CVEs to be fixed.

See report:

[obsidiandynamics-kafdrop3300-2022-08-10-030036.pdf](https://github.com/obsidiandynamics/kafdrop/files/9296247/obsidiandynamics-kafdrop3300-2022-08-10-030036.pdf)

cbl315:
Is there a plan about when to release next versoin? I can help with those security issues and hope those issues could be fixed in next release.

davideicardi:
Any help is appreciated! If someone can create a Pull Request explaining the fix I will be happy to merge it.

Bert-R:
@cbl315 Can you share the list of CVE numbers? Just a Fortify scan report isn't very helpful, as Fortify is known to provide very many false positives. The report (the detailed one, not the high-level summary) needs to be interpreted by one who knows both Fortify and the product (Kafdrop) very well.

cbl315:
> @cbl315 Can you share the list of CVE numbers? Just a Fortify scan report isn't very helpful, as Fortify is known to provide very many false positives. The report (the detailed one, not the high-level summary) needs to be interpreted by one who knows both Fortify and the product (Kafdrop) very well.

Sure I can share the CVEs list, btw I use `protecode` as scan tool instead of `fortify`. Unfortunately the scan result of protecode also contains many false positive vulnerabilities, I might try `trivy` also as a reference. 

[obsidiandynamics-kafdrop3300-vulnerabilities.csv](https://github.com/obsidiandynamics/kafdrop/files/9298129/obsidiandynamics-kafdrop3300-vulnerabilities.csv)

cbl315:
Just a curiosity, which security scan tool is used before for kafdrop?

cbl315:
Update: after I replace the base image to the latest release from upstream, the CVEs number has been reduced from 79 to 41. But there are still donzens, I will keep looking and try to check whether they are false positive.

Base image I use to replace: `eclipse-temurin@sha256:555091411bbe4d768d73b9328b1a62bde263fa36f53f49452e2d92a690eb7a2c`. 
[Here](https://hub.docker.com/layers/eclipse-temurin/library/eclipse-temurin/11/images/sha256-555091411bbe4d768d73b9328b1a62bde263fa36f53f49452e2d92a690eb7a2c?context=explore) is docker hub url.

[...]

PR #404, opened 2022-08-10
feat: enable Dependabot to update deps and update base image tag

Bert-R:
Dependabot will now create pull requests when new Docker images, dependencies or GitHub actions become available. By fixing the base image tag, we always exactly know what base image we are using and Dependabot will propose updates.

This will reduce the probability of accumulating known vulnerabilities, like suggested in #403

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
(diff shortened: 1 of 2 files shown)


davideicardi (approved):
Nice! Thank you!

Bert-R:
@davideicardi It initially gives a barrage of PRs, as you face now. But it'll become manageable after the first wave.

PR #420, opened 2022-08-30
Add option to set topic message retention time in topic-detail page

amtyf:
before-adding-retention
after-adding-retention


davideicardi:
Thank you @amtyf ! LGTM!

@ekoutanov, do you have any feedback?

PR #439, opened 2022-10-19
:seedling: Add hostAliases to the deployment.yaml and values.yaml files

mostafaghadimi:
In this pull request, I have added `hostAliases` in deployment manifest and also put it in values.yaml file.

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -23,6 +23,10 @@ spec:
 {{ toYaml .Values.podAnnotations | indent 8 }}
 {{- end }}        
     spec:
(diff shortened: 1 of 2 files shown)


davideicardi:
Thank you @mostafaghadimi!
Can you just explain what is your scenario? Why you need to modify host aliases?

mostafaghadimi:
Sure @davideicardi,

Occasionally, when there are too many nodes in a Kafka cluster, it is necessary to use something like a DNS server to pass the domain instead of the IP address. This was achieved by adding `hostAliases` to `values.yaml` and deployment template.

PR #470, opened 2023-02-15
Fix js json large bigint display

### LINKED ISSUE #469: JSON parse/stringify with large bigint display incorrectly
reference: https://stackoverflow.com/questions/18755125/node-js-is-there-any-proper-way-to-parse-json-with-large-numbers-long-bigin

IceMimosa:
Fix #469


davideicardi:
Thank you @IceMimosa !
My only concern is that this could potentially break others parts. What is the source of `/js/json-bigint-1.0.0.min.js`?

IceMimosa:
@davideicardi https://github.com/sidorares/json-bigint
I test some production json data locally, and it works fine.

PR #472, opened 2023-02-17
Fix link from partition to detail page when a message.format is set.

Algorab:
If a message format is set, the link from partition to the detail isn't correct. The detail page response with an exception stack trace.


davideicardi:
Do you know why it doesn't work when passing message format?

PR #488, opened 2023-03-21
WIP - Do not Merge: Added support for AWS Glue Schema registry

nicklester:
Added support to enable the AWS Glue Schema Registry

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,12 @@ and if you also require basic auth for your schema registry connection you shoul
 --schemaregistry.auth=username:password
 ```
 
+or, if you are using the AWS Glue Schema Registry
+```
+--schemaregistry.glue.region=us-east-1 --schemaregistry.glue.registryName=demo-registry
+```
+
+
 Finally, a default message and key format (e.g. to deserialize Avro messages or keys) can optionally be configured as follows:
 ```
 --message.format=AVRO
--- a/src/main/java/kafdrop/controller/MessageController.java
+++ b/src/main/java/kafdrop/controller/MessageController.java
@@ -28,6 +28,7 @@ import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
 
+import kafdrop.config.GlueSchemaRegistryConfiguration;
 import kafdrop.util.*;
 import org.springframework.http.MediaType;
 import org.springframework.stereotype.Controller;
@@ -48,6 +49,7 @@ import io.swagger.annotations.ApiResponses;
 import kafdrop.config.MessageFormatConfiguration.MessageFormatProperties;
 import kafdrop.config.ProtobufDescriptorConfiguration.ProtobufDescriptorProperties;
 import kafdrop.config.SchemaRegistryConfiguration.SchemaRegistryProperties;
+import kafdrop.config.GlueSchemaRegistryConfiguration.GlueSchemaRegistryProperties;
 import kafdrop.model.MessageVO;
 import kafdrop.model.TopicPartitionVO;
 import kafdrop.model.TopicVO;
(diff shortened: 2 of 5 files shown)


davideicardi:
Thank you @nicklester !
It LGTM, maybe I would only prefer to have the AWS Glue implementation a little more "isolated" ... What do you think @Bert-R ?

src/main/java/kafdrop/controller/MessageController.java
```diff
@@ -65,14 +67,17 @@ public final class MessageController {
...
   private final ProtobufDescriptorProperties protobufProperties;
 
-  public MessageController(KafkaMonitor kafkaMonitor, MessageInspector messageInspector, MessageFormatProperties messageFormatProperties, SchemaRegistryProperties schemaRegistryProperties, ProtobufDescriptorProperties protobufProperties) {
+  public MessageController(KafkaMonitor kafkaMonitor, MessageInspector messageInspector, MessageFormatProperties messageFormatProperties, SchemaRegistryProperties schemaRegistryProperties, ProtobufDescriptorProperties protobufProperties, GlueSchemaRegistryProperties glueSchemaRegistryProperties) {
     this.kafkaMonitor = kafkaMonitor;
     this.messageInspector = messageInspector;
     this.messageFormatProperties = messageFormatProperties;
     this.schemaRegistryProperties = schemaRegistryProperties;
-    this.protobufProperties = protobufProperties; 
+    this.glueSchemaRegistryProperties = glueSchemaRegistryProperties;
```
Bert-R, line 79: Move this one down, to maintain the same order as the parameter list

README.md
```diff
@@ -60,6 +60,12 @@ and if you also require basic auth for your schema registry connection you shoul
 --schemaregistry.auth=username:password
 ```
 
+or, if you are using the AWS Glue Schema Registry
+```
+--schemaregistry.glue.region=us-east-1 --schemaregistry.glue.registryName=demo-registry
```
Bert-R, line 65:
```suggestion
--schemaregistry.glue.region=us-east-1
--schemaregistry.glue.registryName=demo-registry
```

Besides this, also document the other (apparently optional) properties.

src/main/java/kafdrop/util/AvroMessageDeserializer.java
```diff
@@ -1,28 +1,59 @@
...
+  }
+
+  private static Deserializer getAWSDeserializer(String region, String registryName, String awsEndpoint) {
+    final var config = new HashMap<String, Object>();
+    config.put(AWSSchemaRegistryConstants.AWS_REGION, region);
+    config.put(AWSSchemaRegistryConstants.REGISTRY_NAME, registryName);
+    config.put(AWSSchemaRegistryConstants.AVRO_RECORD_TYPE, AvroRecordType.GENERIC_RECORD.getName());
+    config.put(AWSSchemaRegistryConstants.DATA_FORMAT, DataFormat.AVRO.name());
+    if(StringUtils.isNotEmpty(awsEndpoint))
+      config.put(AWSSchemaRegistryConstants.AWS_ENDPOINT, awsEndpoint);
```
Bert-R, line 51:
```suggestion
    if (StringUtils.isNotEmpty(awsEndpoint)) {
      config.put(AWSSchemaRegistryConstants.AWS_ENDPOINT, awsEndpoint);
    }
```

We'll need to establish and publish a code style, but from what I see, curly braces are used around all blocks.

Bert-R (changes requested):
I've added a few small comments. 
I agree with @davideicardi that a bit more isolation and genericity would be helpful.

What about this:
* Add a property ``schemaregistry.type``, with possible values ``generic`` and ``aws-glue``. Default is ``generic``.
* Extend ``SchemaRegistryProperties`` with a check: if ``type != 'generic'``, then the properties ``connect`` and ``auth`` should not be set.
* Enhance ``GlueSchemaRegistryProperties`` in a similar way and remove ``isConfigured``. Als add a validation that the mandatory properties are set if this registry type is configured.
* Extend ``AvroMessageDeserializer`` with a map of deserializer factories, keyed by the registry type.

With that, we're prepared for a future PR that adds support for  [Microsoft's ``SchemaRegistryApacheAvroSerializer``](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/schemaregistry/azure-data-schemaregistry-apacheavro#create-schemaregistryapacheavroserializer-instance)

src/main/java/kafdrop/util/AvroMessageDeserializer.java
nicklester: good suggestiosn - sorry slow response. Busy work wise. Will take a look shortly

nicklester:
> I've added a few small comments. I agree with @davideicardi that a bit more isolation and genericity would be helpful.
> 
> What about this:
> 
> * Add a property `schemaregistry.type`, with possible values `generic` and `aws-glue`. Default is `generic`.
> * Extend `SchemaRegistryProperties` with a check: if `type != 'generic'`, then the properties `connect` and `auth` should not be set.
> * Enhance `GlueSchemaRegistryProperties` in a similar way and remove `isConfigured`. Als add a validation that the mandatory properties are set if this registry type is configured.
> * Extend `AvroMessageDeserializer` with a map of deserializer factories, keyed by the registry type.
> 
> With that, we're prepared for a future PR that adds support for [Microsoft's `SchemaRegistryApacheAvroSerializer`](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/schemaregistry/azure-data-schemaregistry-apacheavro#create-schemaregistryapacheavroserializer-instance)

Was just a quick question - (as Java is not my 1st or even 2nd language!). Isn't using a factory going to mean we need to pass the same types as arguments, but the generic and glue configs are different, and in different hierarchies). Apologies if I've missed something obvious :)

Bert-R:
No problem! As always, there are multiple ways to approach it. I'd do a slightly bigger refactor as you find in 
[the attached patch](https://github.com/obsidiandynamics/kafdrop/files/11104590/schema-factory.patch). Note that I have not done _any_ testing. The constants for the schema registry type would need to be moved from ``MessageDeserializerFactory`` to a ``SchemaRegistryConfiguration``, to use them to test the possible values for ``type``.

PR #502, opened 2023-04-06
Don't remove KAFKA_PROPERTIES_FILE; allow user to use mounted file

brsolomon-deloitte:
This change gives users the option to put their _own_ `KAFKA_PROPERTIES_FILE` into an image _without_ specifying `KAFKA_PROPERTIES`.

This is particularly useful in cases where a user does not want to store `KAFKA_PROPERTIES` in version control such as Git. (Even though it is base64-encoded, it is not encrypted.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,10 +268,12 @@ EOF
 
 docker run -d --rm -p 9000:9000 \
     -v $(pwd)/kafka.properties:/tmp/kafka.properties:ro \
+    -v $(pwd)/kafka.truststore.jks:/tmp/kafka.truststore.jks:ro \
+    -v $(pwd)/kafka.keystore.jks:/tmp/kafka.keystore.jks:ro \
     -e KAFKA_BROKERCONNECT=<host:port,host:port> \
     -e KAFKA_PROPERTIES_FILE=/tmp/kafka.properties \
-    -e KAFKA_TRUSTSTORE="$(cat kafka.truststore.jks | base64)" \   # optional
-    -e KAFKA_KEYSTORE="$(cat kafka.keystore.jks | base64)" \       # optional
+    -e KAFKA_TRUSTSTORE_FILE=/tmp/kafka.truststore.jks \   # optional
+    -e KAFKA_KEYSTORE_FILE=/tmp/kafka.keystore.jks \       # optional
     obsidiandynamics/kafdrop
 ```
 
--- a/README.md
+++ b/README.md
@@ -256,6 +256,19 @@ docker run -d --rm -p 9000:9000 \
     -e KAFKA_KEYSTORE="$(cat kafka.keystore.jks | base64)" \       # optional
     obsidiandynamics/kafdrop
 ```
+
+Rather than passing `KAFKA_PROPERTIES` as a base64-encoded string, you can also place a pre-populated `KAFKA_PROPERTIES_FILE` into the container:
+
+```sh
+docker run -d --rm -p 9000:9000 \
+    -v $(pwd)/kafka.properties:/tmp/kafka.properties:ro \
+    -e KAFKA_BROKERCONNECT=<host:port,host:port> \
+    -e KAFKA_PROPERTIES_FILE=/tmp/kafka.properties \
+    -e KAFKA_TRUSTSTORE="$(cat kafka.truststore.jks | base64)" \   # optional
+    -e KAFKA_KEYSTORE="$(cat kafka.keystore.jks | base64)" \       # optional
+    obsidiandynamics/kafdrop
+```
+
 #### Environment Variables
 ##### Basic configuration
 |Name                   |Description
--- a/README.md
+++ b/README.md
@@ -260,6 +260,12 @@ docker run -d --rm -p 9000:9000 \
 Rather than passing `KAFKA_PROPERTIES` as a base64-encoded string, you can also place a pre-populated `KAFKA_PROPERTIES_FILE` into the container:
 
 ```sh
+cat << EOF > kafka.properties
+security.protocol=SASL_SSL
+sasl.mechanism=SCRAM-SHA-512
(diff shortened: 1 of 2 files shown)


src/main/docker/kafdrop.sh
```diff
@@ -40,7 +40,7 @@ if [ "$KAFKA_PROPERTIES" != "" ]; then
   echo Writing Kafka properties into $KAFKA_PROPERTIES_FILE
   echo "$KAFKA_PROPERTIES" | base64 --decode --ignore-garbage > $KAFKA_PROPERTIES_FILE
 else
-  rm $KAFKA_PROPERTIES_FILE |& > /dev/null | true
+  if [ -f $KAFKA_PROPERTIES_FILE ]; then touch $KAFKA_PROPERTIES_FILE; else rm -f $KAFKA_PROPERTIES_FILE; fi
```
Bert-R: * Why are you touching the file if it exists?
* Shouldn't you add ``-r`` to the ``rm`` command? If it's not a regular file, it might be a directory.

Bert-R (commented):
Good addition!
Can you update the ``README.md`` and add an alternative ``docker run`` where the Kafka properties file is mounted in read-only?

Something like this:

```
docker run -d --rm -p 9000:9000 \
    -v $(pwd)/kafka.properties:/tmp/kafka.properties:ro \
    -e KAFKA_BROKERCONNECT=<host:port,host:port> \
    -e KAFKA_PROPERTIES_FILE=/tmp/kafka.properties \
    -e KAFKA_TRUSTSTORE="$(cat kafka.truststore.jks | base64)" \   # optional
    -e KAFKA_KEYSTORE="$(cat kafka.keystore.jks | base64)" \       # optional
    obsidiandynamics/kafdrop
```

src/main/docker/kafdrop.sh
brsolomon-deloitte: I've simplified this logic now to 

```bash
if [ ! -f $KAFKA_PROPERTIES_FILE ]; then touch $KAFKA_PROPERTIES_FILE; fi
```

The idea is that, if the default `kafka.properties` is being used, it should be written out as a blank file as Kafdrop will look for it at `KAFKA_PROPERTIES_FILE`.

Whereas, if it already exists (e.g. if the user has made it available through a volume mount), then just let it be.

brsolomon-deloitte:
> Can you update the README.md and add an alternative docker run where the Kafka properties file is mounted in read-only?

Added.

src/main/docker/kafdrop.sh
```diff
@@ -40,7 +40,7 @@ if [ "$KAFKA_PROPERTIES" != "" ]; then
   echo Writing Kafka properties into $KAFKA_PROPERTIES_FILE
   echo "$KAFKA_PROPERTIES" | base64 --decode --ignore-garbage > $KAFKA_PROPERTIES_FILE
 else
-  rm $KAFKA_PROPERTIES_FILE |& > /dev/null | true
+  if [ ! -f $KAFKA_PROPERTIES_FILE ]; then touch $KAFKA_PROPERTIES_FILE; fi
```
Bert-R: Please merge this line with the one above like this:
```suggestion
  elif [ ! -f $KAFKA_PROPERTIES_FILE ]; then
    touch $KAFKA_PROPERTIES_FILE;
```

Can you apply this pattern for the keystore and the truststore too and update the README accordingly? That basically makes this PR equal to #242 (which I now see for the first time), except for the added documentation.

It will also enable running Kafdrop with secrets in Docker compose or K8s.

src/main/docker/kafdrop.sh
Bert-R: That pattern makes sense indeed. @ChristianCiach already [flagged the deletion long back](https://github.com/obsidiandynamics/kafdrop/commit/b07a84d745ffec682f32548ff1e24ab4dbabb8ed#r53695036), I just noticed.

src/main/docker/kafdrop.sh
brsolomon-deloitte: Thanks, updated in [5ffb777](https://github.com/obsidiandynamics/kafdrop/pull/502/commits/5ffb777f027c78fd23450908941da5d2fbeecee9)

README.md
```diff
@@ -256,6 +256,25 @@ docker run -d --rm -p 9000:9000 \
...
+sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="foo" password="bar"
+EOF
+
+docker run -d --rm -p 9000:9000 \
+    -v $(pwd)/kafka.properties:/tmp/kafka.properties:ro \
+    -e KAFKA_BROKERCONNECT=<host:port,host:port> \
+    -e KAFKA_PROPERTIES_FILE=/tmp/kafka.properties \
+    -e KAFKA_TRUSTSTORE="$(cat kafka.truststore.jks | base64)" \   # optional
+    -e KAFKA_KEYSTORE="$(cat kafka.keystore.jks | base64)" \       # optional
+    obsidiandynamics/kafdrop
```
Bert-R:
```suggestion
docker run -d --rm -p 9000:9000 \
    -v $(pwd)/kafka.properties:/tmp/kafka.properties:ro \
    -v $(pwd)/kafka.truststore.jks:/tmp/kafka.truststore.jks:ro \
    -v $(pwd)/kafka.keystore.jks:/tmp/kafka.keystore.jks:ro \
    -e KAFKA_BROKERCONNECT=<host:port,host:port> \
    -e KAFKA_PROPERTIES_FILE=/tmp/kafka.properties \
    -e KAFKA_TRUSTSTORE_FILE=/tmp/kafka.truststore.jks \   # optional
    -e KAFKA_KEYSTORE_FILE=/tmp/kafka.keystore.jks \       # optional
    obsidiandynamics/kafdrop
```
Nearly done. Please update the README to also use the mounted files for the key store and trust store.

Bert-R (approved):
Looks good. Thanks!

PR #506, opened 2023-04-13
fix: message.keyFormat not working

nemesb:
On the messages page the default value for key format should be defaultKeyFormat, not defaultFormat


Bert-R (approved):
Good catch. Thanks!

PR #510, opened 2023-04-18
Ignoring unnecessarily generated docker-ready

optimizing-ci-builds:
In our analysis, we observe that target/docker-ready is generated but not used in CI. Because the contents of docker-ready/ are not accessed after they are generation, we propose to disable the generation of this directory to save build runtime. The generation of this directory can be disabled by simply adding a property element to the pom.xml or adding an argument to the mvn command. The argument to add to the mvn command is -Dskip.docker.resources=true.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -29,4 +29,4 @@ jobs:
         distribution: 'adopt'
         cache: maven
     - name: Build with Maven
-      run: mvn -B verify --file pom.xml
+      run: mvn -B verify --file pom.xml -Dskip.docker.resources=true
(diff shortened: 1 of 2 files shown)


Bert-R:
Though it's true that this folder is not being used during the verification, I don't see any practical benefit in adding this condition, as the performance overhead of copying these two tiny files is negligible.

What gain are you expecting?

Bert-R:
@davideicardi Your thoughts?

davideicardi:
No strong opinion on my side, but 14% time savings is quite good.
If there aren't drawbacks why not! We will save some CO2 :grin:

Bert-R:
@davideicardi Would you mind testing it for yourself? Just checkout this PR and build with ``-Dskip.docker.resources=true`` and ``-Dskip.docker.resources=false`` and compare the times. I didn't notice any difference, so then I felt extra complexity is not justified.

davideicardi:
@Bert-R You are right, I don't see any relevant difference. @optimizing-ci-builds How do you have misured this?

PR #518, opened 2023-05-03
Ignoring unnecessarily generated surefire-report

java-codehunger:
In our analysis, we observe that this target/surefire-report/ directory is generated but not later used during the CI build. Hence, we propose to disable generating this directory to save runtime.
The generation of this directory can be disabled by simply adding -Dsurefire.useFile=false -DdisableXmlReport=true to the mvn command.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -33,4 +33,4 @@ jobs:
         distribution: 'adopt'
         cache: maven
     - name: Build with Maven
-      run: mvn -B verify --file pom.xml
+      run: mvn -B verify --file pom.xml -Dsurefire.useFile=false -DdisableXmlReport=true


.github/workflows/pull_request.yml
```diff
@@ -33,4 +33,4 @@ jobs:
         distribution: 'adopt'
         cache: maven
     - name: Build with Maven
-      run: mvn -B verify --file pom.xml
+      run: mvn -B verify --file pom.xml -Dsurefire.useFile=false -DdisableXmlReport=true
```
Bert-R, line 36: The flag ``disableXmlReport`` is deprecated, see [the documentation](https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#disableXmlReport).

Bert-R (changes requested):
Thanks for the improvement suggestion. However, one of the used flags is deprecated, see my other comment.

PR #524, opened 2023-05-13
[503] Generate new ASCII art for Kafdrop 4

### LINKED ISSUE #503: Generate new ASCII art for Kafdrop 4
Now that the major version number is incremented, we need new ASCII art in [``kafdrop.sh``](https://github.com/obsidiandynamics/kafdrop/blob/master/src/main/docker/kafdrop.sh).

randr97:
This PR Closes #503


Bert-R (approved):
Thanks @randr97 !

Can you share a link to the generator you used? Just in case we ever need to do this again.

PR #537, opened 2023-06-22
non-root user

rseleven:
Running a Container With a Non Root User

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
-        <version>3.1.0</version>
+        <version>3.1.1</version>
     </parent>
     <groupId>com.obsidiandynamics.kafdrop</groupId>
     <artifactId>kafdrop</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -239,9 +239,9 @@
                 </configuration>
             </plugin>
             <plugin>
-                <groupId>com.spotify</groupId>
+                <groupId>io.fabric8</groupId>
                 <artifactId>docker-maven-plugin</artifactId>
-                <version>1.2.2</version>
+                <version>0.43.0</version>
                 <configuration>
                     <imageName>obsidiandynamics/kafdrop</imageName>
                     <forceTags>true</forceTags>
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -1,9 +1,15 @@
 FROM eclipse-temurin:17.0.7_7-jdk
 
-ADD kafdrop.sh /
-ADD kafdrop*tar.gz /
+COPY kafdrop.sh /
+COPY kafdrop*tar.gz /
 
-RUN chmod +x /kafdrop.sh
+RUN chmod +x /kafdrop.sh && \
+    groupadd -g 1000 kafdrop && \
+    useradd -u 1000 -g kafdrop kafdrop
+
+USER kafdrop
+
+ENV PATH="/opt/java/openjdk/bin:$PATH"
 
 EXPOSE 9000
 
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -1,11 +1,20 @@
 FROM eclipse-temurin:17.0.7_7-jdk
 
-COPY kafdrop.sh /
-COPY kafdrop*tar.gz /
+COPY kafdrop*.tar.gz /tmp/kafdrop.tar.gz
 
-RUN chmod +x /kafdrop.sh && \
+RUN mkdir /opt/kafdrop && \
+    tar \
+      -xz \
+      --strip-components=1 \
+      --no-same-owner \
+      -C /opt/kafdrop \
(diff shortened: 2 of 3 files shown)


Bert-R (approved):

davideicardi (approved):

src/main/docker/Dockerfile
```diff
@@ -1,9 +1,15 @@
 FROM eclipse-temurin:17.0.7_7-jdk
 
-ADD kafdrop.sh /
-ADD kafdrop*tar.gz /
+COPY kafdrop.sh /
+COPY kafdrop*tar.gz /
```
Bert-R: This won't work: the ``ADD` command extracts the zip file. Now it's just copied.

Bert-R (changes requested):
The intend is good, but the current approach does not work:
* The content is not added properly (see my other comment)
* Kafdrop does not run anymore. This is because the working directory is set to `/`, which is only writable by root. Due to that, the `touch` operations of `kafdrop.sh` all fail. The right way to fix that is by adding the application in `/app` and make that the working directory. However, that requires changes in `Dockerfile`, `kafdrop.sh` and the README

rseleven:
- replace-dependencies: com.spotify.docker-maven-plugin:1.22 -> io.fabric8.docker-maven-plugin:0.43.0
- fix "Dockerfile"
- fix "kafdrop.sh"

when using com.spotify.docker-maven-plugin:1.22 there were problems with https://docs.docker.com/build/buildkit/

pom.xml
```diff
@@ -239,9 +239,9 @@
                 </configuration>
             </plugin>
             <plugin>
-                <groupId>com.spotify</groupId>
+                <groupId>io.fabric8</groupId>
                 <artifactId>docker-maven-plugin</artifactId>
-                <version>1.2.2</version>
```
Bert-R, line 244: Please revert this change, as the current change is incomplete, leading to these warnings that make the image build fail:
```
[WARNING] Parameter 'dockerDirectory' is unknown for plugin 'docker-maven-plugin:0.43.0:build (default-cli)'
[WARNING] Parameter 'imageName' is unknown for plugin 'docker-maven-plugin:0.43.0:build (default-cli)'
[WARNING] Parameter 'forceTags' is unknown for plugin 'docker-maven-plugin:0.43.0:build (default-cli)'
[WARNING] Parameter 'resources' is unknown for plugin 'docker-maven-plugin:0.43.0:build (default-cli)'
[WARNING] Parameter 'imageTags' is unknown for plugin 'docker-maven-plugin:0.43.0:build (default-cli)'
```

With that, ``pom.xml`` can be removed from your change set.

src/main/docker/Dockerfile
```diff
@@ -1,10 +1,27 @@
 FROM eclipse-temurin:17.0.7_7-jdk
 
-ADD kafdrop.sh /
-ADD kafdrop*tar.gz /
+COPY kafdrop*.tar.gz /tmp/kafdrop.tar.gz
 
-RUN chmod +x /kafdrop.sh
+RUN mkdir /opt/kafdrop && \
+    tar \
```
Bert-R, line 6: This became a lot more complex than it used to be. Let's simplify this.
* Through `ADD`, Docker can extract the zip to the `/opt/kafdrop`. That causes a bit of an ugly directory structure, but nobody needs to go through it, so it doesn't matter
* The `chmod` can be prevented by adding the execution permission in Git. I created #539 for that

With these changes, the `Dockerfile` can be come this simple:
```
FROM eclipse-temurin:17.0.7_7-jdk

WORKDIR /opt/kafdrop
ADD kafdrop*tar.gz .
COPY kafdrop.sh .

RUN groupadd -g 1000 kafdrop && \
    useradd -u 1000 -g kafdrop kafdrop && \
    chown --recursive kafdrop:kafdrop /opt/kafdrop

USER kafdrop

ENV PATH="/opt/java/openjdk/bin:$PATH"

EXPOSE 9000

ENTRYPOINT ["/opt/kafdrop/kafdrop.sh"]
```

src/main/docker/kafdrop.sh
```diff
@@ -65,4 +65,4 @@ ARGS="--add-opens=java.base/sun.nio.ch=ALL-UNNAMED -Xss256K \
      $HEAP_ARGS \
      $JVM_OPTS"
 
-exec java $ARGS -jar /kafdrop*/kafdrop*jar ${CMD_ARGS}
+exec java $ARGS -jar /opt/kafdrop/kafdrop*.jar ${CMD_ARGS}
```
Bert-R, line 68: Due to the above mentioned changes, this path needs to be changed:
```suggestion
exec java $ARGS -jar /opt/kafdrop/kafdrop*/kafdrop*jar ${CMD_ARGS}
```

Bert-R (changes requested):
Thanks for the follow-up!
Unfortunately, the build didn't work anymore because the newly selected plug-in uses different configuration parameters.

I've looked into it and feel we can work with a much simplified `Dockerfile`, while still using the 'old' Spotify Docker plug-in. See my comments.

To test locally, you would need to do a local `chmod +x src/main/docker/kafdrop.sh` until @davideicardi has reviewed and merged #539.

Issue #579, opened 2023-10-09
Kafdrop 4 complains about missing key store password though keystore was not set


mperello-tech:
Hello,

I have tried to install kafdrop4 via helm chart and i left the configs related to key store empty since I only need to use kafka.properties for SASL-SCRAM but on startup an error is thrown as bellow:

```
Caused by: org.apache.kafka.common.KafkaException: org.apache.kafka.common.errors.InvalidConfigurationException: SSL key store is specified, but key store password is not specified.
	at org.apache.kafka.common.network.SaslChannelBuilder.configure(SaslChannelBuilder.java:184)
	at org.apache.kafka.common.network.ChannelBuilders.create(ChannelBuilders.java:192)
	at org.apache.kafka.common.network.ChannelBuilders.clientChannelBuilder(ChannelBuilders.java:81)
	at org.apache.kafka.clients.ClientUtils.createChannelBuilder(ClientUtils.java:105)
	at org.apache.kafka.clients.consumer.KafkaConsumer.<init>(KafkaConsumer.java:737)
	... 36 more
Caused by: org.apache.kafka.common.errors.InvalidConfigurationException: SSL key store is specified, but key store password is not specified.
```

**snippet of values.yaml**
```
image:
  repository: obsidiandynamics/kafdrop
  tag: latest
  pullPolicy: Always

kafka:
  brokerConnect: ${host}:9096
  properties: ""
  truststore: ""
  keystore: ""
  propertiesFile: "/kafka/kafka.properties"
  truststoreFile: ""
  keystoreFile: ""
```

This issue is not encountered with kafdrop 3.31.x

Hope you can help me point to right direction in fixing this.

Many thanks!

davidcurrie:
I believe [this change](https://github.com/obsidiandynamics/kafdrop/commit/d5636069c8702b2a522a6cb9e092bcba30b10243#diff-832ab3de2c8910a0a2d9f4fe4827e9a1ccde40835d416f3d7f89cb806d4226bfL59) is the problem. We now always end up with a keystore file (albeit empty), even when we don't want one.

Bert-R:
That's not the issue. It runs fine with this Docker Compose fragment:

```
 kafdrop:
    container_name: kafdrop
    image: obsidiandynamics/kafdrop:4.0.0
    restart: "no"
    ports:
      - "9000:9000"
    environment:
      KAFKA_BROKERCONNECT: "kafka:29092"
[...]

PR #583, opened 2023-10-27
don't touch keystore or it will try and use it

msillence:
https://github.com/obsidiandynamics/kafdrop/issues/579

### DIFF
--- a/src/main/docker/kafdrop.sh
+++ b/src/main/docker/kafdrop.sh
@@ -36,7 +36,6 @@ if [ $JMX_PORT ]; then
     -Djava.rmi.server.hostname=$HOST"
 fi
 
-KAFKA_PROPERTIES_FILE=${KAFKA_PROPERTIES_FILE:-kafka.properties}
 if [ "$KAFKA_PROPERTIES" != "" ]; then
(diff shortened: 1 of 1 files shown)


msillence:
Looking at the commit that brought this in: https://github.com/obsidiandynamics/kafdrop/commit/d5636069c8702b2a522a6cb9e092bcba30b10243#diff-832ab3de2c8910a0a2d9f4fe4827e9a1ccde40835d416f3d7f89cb806d4226bfL59

I guess we could touch them if the env is set but defaulting it and setting it seems to cause problems
I'm not sure we should

Bert-R:
Sorry for not getting to this earlier, but I don't think this is the right fix. The issue that #502 wanted to address is that the config files were deleted. Instead of just not deleting them, moved to creating them. Now I created #584 to leave the files untouched. If they exist, they won't be deleted. If they don't exist, they won't be created, unless the content of the files is provided through environment variables.

Bert-R:
This is fixed in a slightly different way through #584. It's now available in release [4.0.1](https://github.com/obsidiandynamics/kafdrop/releases/tag/4.0.1)

PR #601, opened 2023-12-01
Fix message formatting when max scan count is exceeded.

monterinio:


src/main/java/kafdrop/service/KafkaMonitorImpl.java
```diff
@@ -280,10 +280,10 @@ public SearchResultsVO searchMessages(String topic,
         case EXCEEDED_MAX_SCAN_COUNT:
           results.setCompletionDetails(
             String.format(
-              "Search timed out after scanning %d messages.  Last scanned message timestamp was %d.  Adjust your time" +
-                " span for more results.",
+              "Search timed out after scanning %d messages. Last scanned message timestamp was %s. Adjust your time" +
+              " span for more results.",
```
Bert-R: Can't we just fix the format?
```suggestion
              "Search timed out after scanning %1$d messages. Last scanned message timestamp was %2$tF %2$tT. Adjust your time" +
                " span for more results.",
```

src/main/java/kafdrop/service/KafkaMonitorImpl.java
```diff
@@ -280,10 +280,10 @@ public SearchResultsVO searchMessages(String topic,
...
               records.getMessagesScannedCount(),
-              records.getFinalMessageTimestamp()));
+              records.finalMessageTimestampFormatted()));
```
Bert-R:
```suggestion
              records.getFinalMessageTimestamp()));
```

Bert-R (changes requested):
Thanks for the contribution!
Added a proposal for a simpler fix.

monterinio:
I will adjust as requested today's evening! Thanks.

src/main/java/kafdrop/service/KafkaMonitorImpl.java
```diff
@@ -280,8 +280,8 @@ public SearchResultsVO searchMessages(String topic,
         case EXCEEDED_MAX_SCAN_COUNT:
           results.setCompletionDetails(
             String.format(
-              "Search timed out after scanning %d messages.  Last scanned message timestamp was %d.  Adjust your time" +
-                " span for more results.",
+              "Search timed out after scanning %d messages. Last scanned message timestamp was %2$tF %2$tT. Adjust your time" +
+              " span for more results.",
```
Bert-R:
```suggestion
                " span for more results.",
```

Bert-R (commented):
Looks good, except for a very small thing: the two spaces on the next line are accidentally removed.

src/main/java/kafdrop/service/KafkaMonitorImpl.java
monterinio: Ha! had the opposite feeling that there are too many of them. (:

Bert-R: The first parameter of `String.format` is broken in two lines. The second half is indented a bit more to show it's not the second parameter but second half of the first one.

Bert-R (approved):
Thanks!

Issue #357, opened 2022-03-10
Publish to artifacthub?


brsolomon-deloitte:
Hi, it'd be great to have this chart officially available on https://artifacthub.io/.

Currently there is https://artifacthub.io/packages/helm/main/kafdrop which is unofficial and points to a very outdated version.

You can use [GitHub pages for the repository](https://medium.com/@mattiaperi/create-a-public-helm-chart-repository-with-github-pages-49b180dbb417) with minimal setup.

davideicardi:
If someone is able to create a PR I think we can publish it during CI pipeline

florianbachmann:
this would be awesome :-)

Issue #236, opened 2021-02-01
Helm Repository


syepes:
It would be great to be able to use your chart directly from the git repo.

Could you consider adding the [index.yaml](https://github.com/cloudworkz/kafka-minion-helm-chart/blob/master/index.yaml) so we could directly do:

`helm repo add kafdrop https://raw.githubusercontent.com/obsidiandynamics/kafdrop/master/chart`

As a reference you can have a look at: https://github.com/cloudworkz/kafka-minion-helm-chart

ekoutanov:
That could work I suppose.

syepes:
I am currently using the minion one and it works like a charm

florianbachmann:
If I try this, I get a 404

```
➜  ~ helm repo add kafdrop https://raw.githubusercontent.com/obsidiandynamics/kafdrop/master/chart
Error: looks like "https://raw.githubusercontent.com/obsidiandynamics/kafdrop/master/chart" is not a valid chart repository or cannot be reached: failed to fetch https://raw.githubusercontent.com/obsidiandynamics/kafdrop/master/chart/index.yaml : 404 Not Found
➜  ~ 

```

can you please help

PR #630, opened 2024-04-03
Add chart-release actions workflow

FieteO:
This PR adds a release workflow as described [on the official helm website](https://helm.sh/docs/howto/chart_releaser_action/#github-actions-workflow).
I can't really test this though, so I am not sure if this works

Related: #236, #357

### DIFF
--- a/.github/workflows/chart-release.yml
+++ b/.github/workflows/chart-release.yml
@@ -1,0 +1,28 @@
+# reference: https://helm.sh/docs/howto/chart_releaser_action/#github-actions-workflow
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


Bert-R (approved):
Thanks for the contribution! Let's try whether it works. I see the installation of Helm is not included, that might cause an issue.

Bert-R:
@FieteO It didn't do anything. This is the output:

```
Looking up latest tag...
Discovering changed charts since '4.0.1'...
Nothing to do. No chart changes detected.
```


I think it's because the folder with the chart is called `/chart` while [the documentation](https://github.com/helm/chart-releaser-action?tab=readme-ov-file#pre-requisites) states it should be `/charts`. Do you mind adding a PR that sets this through the ``charts_dir`` option?

florianbachmann:
Hello hello,
I tried to add the directive here:
https://github.com/obsidiandynamics/kafdrop/pull/631
best regards,
Flori, friend of Fiete :-)

Issue #641, opened 2024-04-30
Consumer Group Id is not html escaped in the UI


davideicardi:
I have received a bug report from "d.morozov" regarding how consumer group id is displayed inside the UI. It looks like it is not html escaped so potentially it can break the UI or we can have have security issues.

Verify if this is true and how we can fix it.

aakashthakare:
Hey @davideicardi,

I tried to investigate this issue and following are my findings,

- For consumer group ID with HTML content is being rendered by UI, html tags can be seen.
- However going into consumer detail page it is failing due to `NoResourceFoundException`
- Escaping HTML from groupId in topic response won't help as `spring.url` will again convert it as unescaped one.
- It would be better to send it in post request rather then get, to solve this problem.

I started looking into the code recently, I could be wrong with my analysis. Feel free to correct me and guide for the same.

Looking forward to contribute more with this project. Let me know if any further investigation is required.

<img width="391" alt="Screenshot 2024-05-19 at 7 36 49 PM" src="https://github.com/obsidiandynamics/kafdrop/assets/97438196/022eab49-80a5-4b00-ac2b-fc0508e98c4d">
<img width="1113" alt="Screenshot 2024-05-19 at 7 54 05 PM" src="https://github.com/obsidiandynamics/kafdrop/assets/97438196/c17a763a-0422-42f0-90fb-bd1e99ddf291">

davideicardi:
Thank you @aakashthakare!
Regarding the resource not found, it is a problem of the id pased in the wrong way? Or maybe some other server side problems?

aakashthakare:
Hello @davideicardi, thanks for the response. Yes, need to fix from server side. The URL `consumer/my-second-consumer-group-<a>new</a>` failing to reach the controller, We need to HTML encode it in URL, while on the server side we need to HTML decode the consumer id.

davideicardi:
What if you pass it as URL encoded? (not HTML encoded ...)

aakashthakare:
It will solve upto some extent, not completely. 

For example, consumer group id is `test-<a>it</a>`, if URL encoded will become `test-%3Ca%3Eit%3C%2Fa%3E`, this we can do while returning the consumer group. However, it gets automatically decoded when clicked on consumer link `consumer/test-%3Ca%3Eit%3C%2Fa%3E`, and we get `A 404 error has occurred: No static resource consumer/test-<a>it</a>`. 

[...]

PR #643, opened 2024-05-22
GroupId Link 404 Fix

aakashthakare:
- Introduce base64 groupId.
- Use encoded groupId in link.
-  Fix 404 error incase groupd id contains HTML tags.

Relates to #641

### DIFF
--- a/src/main/resources/templates/topic-detail.ftlh
+++ b/src/main/resources/templates/topic-detail.ftlh
@@ -156,7 +156,7 @@
                 <tbody>
                 <#list consumers![] as c>
                     <tr>
-                        <td><a href="<@spring.url '/consumer/${c.groupId}'/>">${c.groupId}</a></td>
+                        <td><a href="<@spring.url '/consumer/${c.groupIdBase64}'/>">${c.groupId}</a></td>
                         <td>${c.getTopic(topic.name).lag}</td>
                     </tr>
                 </#list>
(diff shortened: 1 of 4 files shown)


Bert-R:
@davideicardi You are familiar with the issue (#641). Can you review this PR?

src/main/resources/templates/topic-detail.ftlh
```diff
@@ -156,7 +156,7 @@
                 <tbody>
                 <#list consumers![] as c>
                     <tr>
-                        <td><a href="<@spring.url '/consumer/${c.groupId}'/>">${c.groupId}</a></td>
+                        <td><a href="<@spring.url '/consumer/${c.groupIdBase64}'/>">${c.groupId}</a></td>
```
davideicardi, line 159: Instead of using a base64 encoding, why not just url encode it? In this way it should be handled automatically by spring I suppose. Probably you can use `UriComponentsBuilder`. In this way you don't need to change `src/main/java/kafdrop/controller/ConsumerController.java`.

davideicardi (changes requested):
Not sure about base64 encoding. See comment.

src/main/resources/templates/topic-detail.ftlh
aakashthakare: I tried that way as well, however `spring.url` was again converting it to decoded value, which was ending up with error when you click on consumer group link.

PR #653, opened 2024-06-19
Update maven wrapper to 3.9.8

jacobgarbercalian:
The current maven wrapper is version 3.6.1 and fails to build the project with the following error:

```
The plugin org.apache.maven.plugins:maven-compiler-plugin:3.13.0 requires Maven version 3.6.3
```

After running `./mvnw wrapper:wrapper -Dmaven=3.9.8` the build works.


Bert-R (approved):
@jacobgarbercalian Thanks!
@davideicardi I have compared the scripts with what's generated for 3.9.8 and they are exactly identical, so I'm going to approve the workflow run and will merge the PR when the build is successful.

PR #708, opened 2024-12-10
Add support for truststore and keystore password parameters in Kafdrop

wkd-woo:
## Description

This PR introduces support for specifying passwords for the truststore and keystore used in SSL communication with Kafka brokers in Kafdrop. 

Previously, while the truststore file location could be specified using the `KAFKA_TRUSTSTORE_FILE` parameter, there was no way to provide the password required to access it. 

This enhancement addresses that limitation.

refer: https://docs.oracle.com/javadb/10.8.3.0/adminguide/cadminsslclient.html

## Key changes include

1. New Configuration Parameters

- Added `KAFKA_TRUSTSTORE_PASSWORD` to specify the truststore password.
- Added `KAFKA_KEYSTORE_PASSWORD` to specify the keystore password.

2. Code Changes

- Modified `KafkaConfiguration.java` to handle these new parameters and set `ssl.truststore.password` and `ssl.keystore.password` in the Kafka properties if provided.
- Ensured that these parameters are optional, maintaining backward compatibility for users who do not require truststore or keystore passwords.

3. Configuration File Updates

- Updated application.yml to include placeholders for the new parameters, making them configurable via environment variables:
    - `KAFKA_TRUSTSTORE_PASSWORD`
    - `KAFKA_KEYSTORE_PASSWORD`

### DIFF
--- a/src/main/java/kafdrop/config/KafkaConfiguration.java
+++ b/src/main/java/kafdrop/config/KafkaConfiguration.java
@@ -29,8 +29,10 @@ public final class KafkaConfiguration {
   private String saslMechanism;
   private String securityProtocol;
   private String truststoreFile;
+  private String truststorePassword;
   private String propertiesFile;
   private String keystoreFile;
+  private String keystorePassword;
 
   public void applyCommon(Properties properties) {
     properties.setProperty(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, brokerConnect);
@@ -43,12 +45,18 @@ public final class KafkaConfiguration {
     if (new File(truststoreFile).isFile()) {
       LOG.info("Assigning truststore location to {}", truststoreFile);
       properties.put("ssl.truststore.location", truststoreFile);
+      if (truststorePassword != null && !truststorePassword.isEmpty()) {
+        properties.put("ssl.truststore.password", truststorePassword);
+      }
     }
 
     LOG.info("Checking keystore file {}", keystoreFile);
     if (new File(keystoreFile).isFile()) {
       LOG.info("Assigning keystore location to {}", keystoreFile);
       properties.put("ssl.keystore.location", keystoreFile);
+      if (keystorePassword != null && !keystorePassword.isEmpty()) {
+        properties.put("ssl.keystore.password", keystorePassword);
+      }
     }
 
     LOG.info("Checking properties file {}", propertiesFile);
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -41,5 +41,7 @@ kafka:
   saslMechanism: "PLAIN"
   securityProtocol: "SASL_PLAINTEXT"
   truststoreFile: "${KAFKA_TRUSTSTORE_FILE:kafka.truststore.jks}"
-  propertiesFile : "${KAFKA_PROPERTIES_FILE:kafka.properties}"
   keystoreFile: "${KAFKA_KEYSTORE_FILE:kafka.keystore.jks}"
+  propertiesFile: "${KAFKA_PROPERTIES_FILE:kafka.properties}"
+  truststorePassword: "${KAFKA_TRUSTSTORE_PASSWORD}"
+  keystorePassword: "${KAFKA_KEYSTORE_PASSWORD}"


Bert-R:
I do not understand the objective of this PR. As mentioned in the README, you can use `kafka.properties` for configuration information "including key/truststore passwords".

What do you want to add to that?

wkd-woo:
@Bert-R 
As mentioned in the README, I have already tried specifying truststore.password in kafka.properties, but Kafdrop still couldn't connect to our TLS-enabled Kafka brokers. This issue persisted even though the kafka.properties file was correctly mounted as a volume.

Additionally, in the company I work for, security policies prohibit storing passwords in configuration files like properties. (Honestly, I find this requirement quite perplexing myself.) However, storing passwords in objects such as Kubernetes Secrets is allowed.

This is why I created this PR. I believe that passwords required for certificate access should also be managed as separate parameters.

Bert-R:
In that case, I suggest a much simpler implementation. We already support `SSL_KEY_STORE_PASSWORD`. If you [search for that string](https://github.com/search?q=repo%3Aobsidiandynamics%2Fkafdrop%20SSL_KEY_STORE_PASSWORD&type=code), you'll find two occurrences:
* https://github.com/obsidiandynamics/kafdrop/blob/87c6f77dddf1d0fb43c2e13b7039f5f784256442/README.md?plain=1#L338
* https://github.com/obsidiandynamics/kafdrop/blob/87c6f77dddf1d0fb43c2e13b7039f5f784256442/src/main/resources/application.yml#L12

If you copy these lines and adapt them for the  trust store, you're all set.

wkd-woo:
Thank you for your suggestion, @Bert-R 

I understand that we already have `SSL_KEY_STORE_PASSWORD` for the keystore password. 
However, keystore and truststore serve different purposes in TLS communication. 
While the keystore is used for server-side authentication, the truststore is necessary for verifying the broker’s certificate in client-side authentication.

Currently, there is no dedicated parameter for truststore.password, which is required for TLS connections with Kafka brokers. Simply duplicating `SSL_KEY_STORE_PASSWORD` would not solve the issue, as it only addresses the keystore, not the truststore.

This PR proposes adding new parameters (`SSL_TRUST_STORE` and `SSL_TRUST_STORE_PASSWORD`) to handle truststore configuration separately, ensuring proper TLS support.

I hope this clarifies the need for the PR. Please review my PR again.

Bert-R:
Your PR adds `KAFKA_TRUSTSTORE_PASSWORD` and `KAFKA_KEYSTORE_PASSWORD`, right? The latter one, we already have. It's named `SSL_KEY_STORE_PASSWORD`. So there is no need to add that. The only thing that is missing is the trust store password.

We do not have the `SSL_TRUST_STORE_PASSWORD` yet, but that could be added with this simple change:
```patch
diff --git a/README.md b/README.md
index e8e3b41..e4fc3a5 100644
--- a/README.md
+++ b/README.md
@@ -353,6 +353,7 @@ docker run -d --rm -p 9000:9000 \
 | `SSL_KEY_STORE_TYPE`     | Type of SSL keystore. Default is `PKCS12`
 | `SSL_KEY_STORE`          | Path to keystore file
 | `SSL_KEY_STORE_PASSWORD` | Keystore password
+| `SSL_TRUST_STORE_PASSWORD` | Truststore password
 | `SSL_KEY_ALIAS`          | Key alias

 ### Using Helm
diff --git a/src/main/resources/application.yml b/src/main/resources/application.yml
index d80ad48..d52ec18 100644
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -10,6 +10,7 @@ server:
     key-store-type: ${SSL_KEY_STORE_TYPE:PKCS12}
     key-store: ${SSL_KEY_STORE:}
     key-store-password: ${SSL_KEY_STORE_PASSWORD:}
+    trust-store-password: ${SSL_TRUST_STORE_PASSWORD:}
     key-alias: ${SSL_KEY_ALIAS:}
     enabled: ${SSL_ENABLED:false}
```

That's what I meant with "copy these lines and adapt them for the trust store".
Makes sense?

PR #716, opened 2025-01-06
Added IntDeserializer to KeyFormat

DesineSperare:
Added IntDeserializer to KeyFormat to display int keys correctly


Bert-R (approved):
Looks good. Thanks for the contribution!

PR #717, opened 2025-01-10
JSON APIs and bug fixes

ndanilin:
Bug fixes:
1. Removed duplication of messages from `viewAllMessages` (I think it's a critical bug).
2. Added filter by timestamp in `searchRecords`. Search messages by timestamp didn't work because `seekToTimestamp` can't guarantee needed result.

New features:
1. Added JSON API `getAllMessages` (equivalent of `viewAllMessages` that returns JSON array of messages).
2. Added JSON API `searchMessages` (equivalent of `searchMessageForm` that returns JSON array of messages). Here I made custom `SearchMessageFormForJson` and now all fields of form are not required. Also added here search by keys (It might overload this method, but seems very helpful for guaranteed results)


src/main/java/kafdrop/controller/MessageController.java
```diff
@@ -32,49 +34,23 @@
 import kafdrop.config.ProtobufDescriptorConfiguration.ProtobufDescriptorProperties;
 import kafdrop.config.SchemaRegistryConfiguration.SchemaRegistryProperties;
 import kafdrop.form.SearchMessageForm;
-import kafdrop.model.MessageVO;
-import kafdrop.model.TopicPartitionVO;
-import kafdrop.model.TopicVO;
+import kafdrop.form.SearchMessageFormForJson;
+import kafdrop.model.*;
```
Bert-R: In our coding convention, we don't use wildcard imports. Please restore the original specific imports.

src/main/java/kafdrop/controller/MessageController.java
```diff
@@ -32,49 +34,23 @@
...
-import kafdrop.util.ProtobufSchemaRegistryMessageDeserializer;
-
-import java.io.File;
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.List;
-
-
-import kafdrop.util.Serializers;
+import kafdrop.util.*;
```
Bert-R: Same here

src/main/java/kafdrop/controller/MessageController.java
```diff
@@ -32,49 +34,23 @@
...
-import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.*;
+import org.springframework.web.server.ResponseStatusException;
 
-import java.util.Date;
-
-import org.springframework.web.bind.annotation.PostMapping;
-import kafdrop.model.CreateMessageVO;
+import java.io.File;
+import java.util.*;
```
Bert-R: And here

src/main/java/kafdrop/controller/MessageController.java
```diff
@@ -32,49 +34,23 @@
...
 import org.springframework.http.MediaType;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
 import org.springframework.validation.BindingResult;
-import org.springframework.web.bind.annotation.GetMapping;
-import org.springframework.web.bind.annotation.ModelAttribute;
-import org.springframework.web.bind.annotation.PathVariable;
-import org.springframework.web.bind.annotation.RequestParam;
-import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.*;
```
Bert-R: And here

src/main/java/kafdrop/controller/MessageController.java
```diff
@@ -143,6 +119,47 @@ public String viewAllMessages(@PathVariable("name") String topicName,
...
+
+    for (TopicPartitionVO partition : topic.getPartitions()) {
+      messages.addAll(messageInspector.getMessages(topicName,
+        partition.getId(),
+        partition.getFirstOffset(),
+        size,
+        deserializers));
+    }
+
+    messages.sort(Comparator.comparing(MessageVO::getTimestamp));
```
Bert-R: This duplicates the lines 102-116 of `viewAllMessages`. Extract these into a private method and call that method here instead of duplicating them.

src/main/java/kafdrop/controller/MessageController.java
```diff
@@ -328,6 +345,67 @@ public String searchMessageForm(@PathVariable("name") String topicName,
...
+        protobufProperties.getParseAnyProto())
+    );
+
+    var searchResultsVO = kafkaMonitor.searchMessages(
+      topicName,
+      searchMessageForm.getSearchText(),
+      searchMessageForm.getPartition(),
+      searchMessageForm.getMaximumCount(),
+      searchMessageForm.getStartTimestamp(),
+      deserializers);
```
Bert-R: These lines duplicate the lines 325-339 of `searchMessageForm`, except that `descFile` and `msgTypeName` are set to `null`. Extract the lines into a method that takes `descFile` and `msgTypeName` as parameters and call the method here and above.

src/main/java/kafdrop/model/MessageVO.java
```diff
@@ -28,6 +30,7 @@ public final class MessageVO {
   private String message;
   private String key;
   private Map<String, String> headers;
+  @JsonFormat(pattern = "yyyy-MM-dd HH:mm:ss.SSS")
```
Bert-R: Why this format instead of ISO 8601?

Bert-R (changes requested):
@ndanilin Thank you for this PR and sorry for the long delay in reviewing it.
It generally looks good. I have added a few comments regarding imports and code duplication. After these are resolved, the PR can be merged. Then we'll bring a release of Kafdrop.

Bert-R:
@ndanilin Thanks for the import rearrangement! Any chance you get to addressing the other comments soon?

src/main/java/kafdrop/form/SearchMessageFormForJson.java
```diff
@@ -0,0 +1,107 @@
...
+  private Integer maximumCount;
+  @Schema(example = "-1")
+  private Integer partition;
+  @Schema(example = "DEFAULT")
+  private MessageFormat format;
+  @Schema(example = "DEFAULT")
+  private MessageFormat keyFormat;
+  @Schema(type = "string", example = "1970-01-01 03:00:00.000")
+  @DateTimeFormat(pattern = "yyyy-MM-dd HH:mm:ss.SSS")
+  @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd HH:mm:ss.SSS")
```
Bert-R: @ndanilin Why this format instead of the regular ISO 8601 format?

Bert-R:
@ndanilin I done the deduplication that I asked for earlier. Can you have a look to see whether the API still matches your expectations?

ndanilin:
@Bert-R I'm really sorry that I forgot to push the second part of my changes about code duplication. Thank you so much for your help.
Yes, the API works fine.

src/main/java/kafdrop/form/SearchMessageFormForJson.java
ndanilin: @Bert-R this format was for "SearchMessageForm" before my changes (ed9d1cd2). I decided to do the same thing. Should we change it to ISO 8601?

src/main/java/kafdrop/model/MessageVO.java
ndanilin: @Bert-R it's the same as for "SearchMessageForm" and "SearchMessageFormForJson" (my comment below)

src/main/java/kafdrop/form/SearchMessageFormForJson.java
Bert-R: Yes, please change it to the ISO format. That's a better fit for an API.

ndanilin: @Bert-R 
fixed this moment:
1. APIs work with iso format (input and output)
2. UI still works with "yyyy-MM-dd HH:mm:ss.SSS" (haven't dealt with ftlh files before it, so check please)

Bert-R: @ndanilin Sorry, that's not what I meant. My comment only applied to the JSON API. Just this change should suffice:

```diff
- @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd HH:mm:ss.SSS")
+ @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", timezone = "UTC")
```

ndanilin: @Bert-R sorry to complicate your idea.
1. canceled a previous commit
2. replaced the "searchMessages" API with the POST method (because GET doesn't work with @Json format)
3. corrected documentation for "searchMessages"
4. changed @JsonFormat to "SearchMessageFormForJson" and "MessageVO", as you said

Bert-R (approved):
I've corrected the imports and wrapped the lines that were longer than our rules. With that the change is OK.

Thanks for the contribution!

PR #759, opened 2025-06-25
Add systemd unit as user contrib

samveen:
Reduce the time to deploy by contributing my starter systemd unit and associated install instructions, to save new users the time of building up everything from scratch.

Note: This could probably be included as common underlying sources  if, as and when kafdrop gets included into packaging systems by linux distributions further downstream. I could also contribute packaging information for `rpm` and `deb` formats, if that's something that'll be useful.


Bert-R:
@davideicardi  Are you familiar with this matter and able to review this PR?

davideicardi:
@samveen Thank you!

It looks a good contribution to me. Just some small suggestions:
- Rename `contrib/systemd/README.syste.md` to `contrib/systemd/README.md` 
- Add a small introduction regarding systemd and when to use it. Maybe something similar to what you have written inside the PR description.
- In which Linux distribution do you have tested it? Maybe it could be useful to specify it and explain that in other distributions some changes could be necessary

The validate action is failing, but it should be something else. I will investigate.

davideicardi (changes requested):
Thank you! See my comment.

samveen:
@davideicardi  Please review.

>     * Rename `contrib/systemd/README.syste.md` to `contrib/systemd/README.md`

Done (even though it takes away that small bit of pun fun in the name).

>     * Add a small introduction regarding systemd and when to use it. Maybe something similar to what you have written inside the PR description.
>     * In which Linux distribution do you have tested it? Maybe it could be useful to specify it and explain that in other distributions some changes could be necessary

Added to relevent sections in README.md

> The validate action is failing, but it should be something else. I will investigate.

The linter failure was due to long lines in the README.md . It's fixed.

davideicardi:
@samveen 
Ah sorry, I just understand now the name joke :-) ! 
In this case we can leave the original name if you prefer!

davideicardi (approved):
Thank you!

samveen:
> @samveen Ah sorry, I just understand now the name joke :-) ! In this case we can leave the original name if you prefer!

I'll leave it as is. The auto-magic display of `README.md` on the Github web interface  is better than making a joke with the name `README.syste.md`. Not to mention removing the need for another round of approvals :rofl:

PR #703, opened 2024-11-25
Added functionality to allow user to add a message to a topic partition manually

nguyen-tri-nhan:
Resolve conflict for #230 

image
image
image

PR #761, opened 2025-07-20
Added feature flag for send message

### LINKED ISSUE #756: Feature Request: Add CLI arguments to disable creating messages
#703 added the ability to produce arbitrary Kafka messages, which is really useful as a debugging tool. Given the power of this though, it would be useful if we could lock it behind a permissions check or command line argument (similar to the topic creation/deletion controls mentioned on <https://github.com/obsidiandynamics/kafdrop?tab=readme-ov-file#topic-configuration>). Perhaps something like `--topic.sendEnabled=false`?

Thank you for the great tool!

nguyen-tri-nhan:
Resolve: #756 
Follow up for: #703 

Create message feat is disabled by default

Set `--message.sendEnabled=true` if we want to use it

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ By default, you could create a topic. If you don't want this feature, you could 
 ```
 
 ## Message Configuration
-By default, you couldn't send messages to a topic. If you want to enable this feature, you could enable it with:
+By default, you cannot send messages to a topic. You can enable it with:
 
 ```
 --message.sendEnabled=true
--- a/README.md
+++ b/README.md
@@ -235,6 +235,13 @@ By default, you could create a topic. If you don't want this feature, you could 
 --topic.createEnabled=false
 ```
 
+## Message Configuration
+By default, you couldn't send messages to a topic. If you want to enable this feature, you could enable it with:
+
+```
+--message.sendEnabled=true
+```
+
 ## Actuator
 Health and info endpoints are available at the following path: `/actuator`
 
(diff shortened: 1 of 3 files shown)


nguyen-tri-nhan:
@Bert-R can you please take a look, It disable the send message feature by default and make people feel confident to use the new version

README.md
```diff
@@ -235,6 +235,13 @@ By default, you could create a topic. If you don't want this feature, you could
 --topic.createEnabled=false
 ```
 
+## Message Configuration
+By default, you couldn't send messages to a topic. If you want to enable this feature, you could enable it with:
```
Bert-R:
```suggestion
By default, you cannot send messages to a topic. You can enable it with:
```

src/main/java/kafdrop/controller/MessageController.java
```diff
@@ -249,6 +255,10 @@ public String addMessage(
     String topicName,
     @ModelAttribute("addMessageForm") CreateMessageVO body,
     Model model) {
+    if (!sendEnabled) {
+      model.addAttribute("errorMessage", "Message sending is not enabled.");
+      return "message-inspector";
+    }
     try {
```
Bert-R: To prevent this early return with the repeated return value, I suggest this approach:
```suggestion
    try {
      if (!sendEnabled) {
        throw new IllegalStateException("Message sending is not enabled.");
      }
```

Bert-R (changes requested):
Thanks for this contribution! It was on my list to implement prior to releasing a new Kafdrop.
I've added two review comments. Please have a look.

README.md
nguyen-tri-nhan: I'm addressed it

src/main/java/kafdrop/controller/MessageController.java
nguyen-tri-nhan: I'm addressed it

Bert-R (approved):
That looks good. Thank you!

PR #763, opened 2025-07-24
Switch to minor release snapshot

Bert-R:
Given that we received some functional enhancements, it would be best to make the next release a minor one rather than a patch


Bert-R:
@davideicardi See my proposal. If you agree, can you merge this PR and then run the regular release steps to deliver this to the interested people?

davideicardi (approved):

PR #778, opened 2025-09-18
Fix crash when rendering non-Date timestamp in search result

vbarjatiya:
Fixes a bug where `String.format("%2$tF %2$tT", timestamp)` fails with `IllegalFormatConversionException` if the timestamp is not a java.util.Date.

This patch replaces the format string with `%s` and uses `String.valueOf(...)` to safely convert the value regardless of its type (Date, Long, String).

### DIFF
--- a/src/main/java/kafdrop/service/KafkaMonitorImpl.java
+++ b/src/main/java/kafdrop/service/KafkaMonitorImpl.java
@@ -287,12 +287,12 @@ public final class KafkaMonitorImpl implements KafkaMonitor {
             "Scanned %d messages.", records.getMessagesScannedCount()));
           break;
         case EXCEEDED_MAX_SCAN_COUNT:
-          results.setCompletionDetails(
-            String.format(
-              "Search timed out after scanning %d messages. Last scanned message timestamp was %2$tF %2$tT." +
-                " Adjust your time span for more results.",
-              records.getMessagesScannedCount(),
-              records.getFinalMessageTimestamp()));
+        results.setCompletionDetails(
+          String.format(
+            "Search timed out after scanning %d messages. Last scanned message timestamp was %s." +
+              " Adjust your time span for more results.",
+            records.getMessagesScannedCount(),
+            String.valueOf(records.getFinalMessageTimestamp())));
           break;
         case NO_MORE_MESSAGES_IN_TOPIC:
           results.setCompletionDetails(


Bert-R:
Thanks for your contribution!
Can you share the stack trace of this failure?

vbarjatiya:
Hello @Bert-R,

Here is the stack trace:

Request processing failed: java.util.IllegalFormatConversionException: d != java.util.Date
	at java.util.Formatter$FormatSpecifier.failConversion(Formatter.java:4442)
	at java.util.Formatter$FormatSpecifier.printInteger(Formatter.java:2963)
	at java.util.Formatter$FormatSpecifier.print(Formatter.java:2918)
	at java.util.Formatter.format(Formatter.java:2689)
	at java.lang.String.format(String.java:4143)
	at kafdrop.service.KafkaMonitorImpl.searchMessages(KafkaMonitorImpl.java:282)
	at kafdrop.controller.MessageController.searchMessageForm(MessageController.java:247)
	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:205)
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:150)
	at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:118)
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:884)
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.handleInternal(RequestMappingHandlerAdapter.java:797)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1081)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:974)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1011)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:527)
	at io.undertow.servlet.handlers.ServletInitialHandler.handleRequest(ServletInitialHandler.java:101)
	at io.undertow.server.Connectors.executeRootHandler(Connectors.java:393)
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2513)
	at java.lang.Thread.run(Thread.java:833)

Bert-R:
I"m puzzled. Your stack trace indicate that a `java.util.Date` value was provided for the `%d` format string. I don't understand how that can happen, nor how the change that you propose would fix that.

Can you dig a bit further and see what values are passed to `String.format` in your case?

vbarjatiya:
Hi @Bert-R,

You are right that the exception message says %d != java.util.Date, which would normally imply a Date was passed to a numeric format placeholder. However, in this specific case, the original Kafdrop code used:
String.format("... %2$tF %2$tT", messageCount, records.getFinalMessageTimestamp())

This format specifier expects records.getFinalMessageTimestamp() to be a java.util.Date. In most cases it works, but we observed in our lab and prod environment that the method sometimes returns a Long or an incompatible object, likely due to how the timestamp is recorded during message search completion

When that happens, Java throws - IllegalFormatConversionException: d != java.util.Date
So the %2$tF formatting fails if the second argument is not a Date. While the exception message mentions a mismatch with %d, the underlying issue is that Java is trying to treat a non-Date value as a timestamp.

We tested this across multiple environments and replacing the format string with:
"%s", String.valueOf(...)
ensures consistent rendering with no crashes regardless of the actual type.

Bert-R:
@vbarjatiya I do not deny that the issue in your environment disappears when you make this change, but I strongly doubt whether it is the right solution.

Here is a part of your stack trace, hyperlinked to the JDK source code:

* `at java.util.Formatter$FormatSpecifier.failConversion`([Formatter.java:4442](https://github.com/openjdk/jdk17/blob/master/src/java.base/share/classes/java/util/Formatter.java#L4442))
* `at java.util.Formatter$FormatSpecifier.printInteger`([Formatter.java:2963](https://github.com/openjdk/jdk17/blob/master/src/java.base/share/classes/java/util/Formatter.java#L2963))
* `at java.util.Formatter$FormatSpecifier.print`([Formatter.java:2918](https://github.com/openjdk/jdk17/blob/master/src/java.base/share/classes/java/util/Formatter.java#L2918))
* `at java.util.Formatter.format`([Formatter.java:2689](https://github.com/openjdk/jdk17/blob/master/src/java.base/share/classes/java/util/Formatter.java#L2689))

This is very clearly the path that formats a `%d`. In your case, it apparently has a `java.util.Date` while it is supposed to be a `float`. Apparently, the stack is messed up. The change that you propose changes the memory layout but it does not fix the real issue.

PR #814, opened 2026-01-07
Handle exception while determining isController flag in KafkaMonitorImpl

deneskelemen:
Sometimes, the clusterDescription.controller is null, and in such cases, the Cluster Overview does not appear.

We have tested this in several of our environments. We have not found any issues that this could have caused.

### DIFF
--- a/src/main/java/kafdrop/service/KafkaMonitorImpl.java
+++ b/src/main/java/kafdrop/service/KafkaMonitorImpl.java
@@ -81,7 +81,13 @@ public final class KafkaMonitorImpl implements KafkaMonitor {
     final var clusterDescription = highLevelAdminClient.describeCluster();
     final var brokerVos = new ArrayList<BrokerVO>(clusterDescription.nodes.size());
     for (var node : clusterDescription.nodes) {
-      final var isController = node.id() == clusterDescription.controller.id();
+      boolean isController;
+      try {
+        isController = node.id() == clusterDescription.controller.id();
+      } catch (Exception e) {
+        LOG.error("Failed to determine if node is controller",e);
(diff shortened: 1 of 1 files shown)


src/main/java/kafdrop/service/KafkaMonitorImpl.java
```diff
@@ -81,7 +81,13 @@ public List<BrokerVO> getBrokers() {
...
     final var brokerVos = new ArrayList<BrokerVO>(clusterDescription.nodes.size());
     for (var node : clusterDescription.nodes) {
-      final var isController = node.id() == clusterDescription.controller.id();
+      boolean isController;
+      try {
+        isController = node.id() == clusterDescription.controller.id();
+      } catch (Exception e) {
+        LOG.error("Failed to determine if node is controller",e);
+        isController = false;
+      }
```
Bert-R, line 90: For maintainability, please extract the code into a separate method that is called here:
```suggestion
      boolean isController = isController(node.id, clusterDescription);
```

In the method, check the controller rather than catching the NPE:
```
  private static boolean isController(int nodeId, DescribeClusterResult clusterDescription)
  {
    var controller = clusterDescription.controller;
    if (controller == null)
    {
      LOG.error("Controller ID not known yet");
      return false;
    }
    return nodeId == controller.id();
  }
```

Bert-R (changes requested):
Good point! According to [the documentation](https://kafka.apache.org/23/javadoc/org/apache/kafka/clients/admin/DescribeClusterResult.html#controller--), the controller yield null, if the controller ID is not yet known.

I've added a comment with a request to improve the code quality.

PR #864, opened 2026-07-30
Show highest last offset and its partition on topic detail

vokracko:
Partition table already lists the offsets, this saves eyeballing the column to find the head of the topic. 
Empty topic renders as `-`.

kafdrop-overview-empty
kafdrop-overview-table


Bert-R:
@vokracko Thanks for your contribution! Copilot spotted something that I missed. Can you apply that little update?

vokracko:
@Bert-R I've accepted the proposed change from copilot. Let me know if there is anything else

Bert-R (approved):
That's it. Thanks!

PR #868, opened 2026-08-09
fix: make one offset request

c4181:
On large clusters, the page for viewing a single topic loads excruciatingly slow. 

Currently, Kafdrop make a separate request to get the offsets for each consumer group in the cluster when the page loads. This PR changes it to make one request for all the groups as recommended by the admin client docs.

> The operation method's first parameter is a Collection of items to perform the operation on. Batching multiple requests into a single call is more efficient and should be preferred over multiple calls to the same method.

### DIFF
--- a/src/main/java/kafdrop/service/KafkaHighLevelAdminClient.java
+++ b/src/main/java/kafdrop/service/KafkaHighLevelAdminClient.java
@@ -6,6 +6,7 @@ import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.Config;
 import org.apache.kafka.clients.admin.ConsumerGroupListing;
 import org.apache.kafka.clients.admin.DeleteTopicsOptions;
+import org.apache.kafka.clients.admin.ListConsumerGroupOffsetsSpec;
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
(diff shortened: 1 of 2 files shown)


Bert-R:
@c4181 Though I doubt whether Kafdrop handles thread interrupts properly, I favor addressing the issue flagged by Copilot, as it's very straightforward.

src/main/java/kafdrop/service/KafkaHighLevelAdminClient.java
```diff
@@ -107,6 +109,31 @@ Map<TopicPartition, OffsetAndMetadata> listConsumerGroupOffsetsIfAuthorized(Stri
...
+      try {
+        offsetsByGroup.put(groupId, result.partitionsToOffsetAndMetadata(groupId).get());
+      } catch (InterruptedException | ExecutionException e) {
+        if (e.getCause() instanceof GroupAuthorizationException) {
+          LOG.info("Not authorized to view consumer group {}; skipping", groupId);
+          offsetsByGroup.put(groupId, Collections.emptyMap());
+        } else {
+          throw new KafkaAdminClientException(e);
+        }
+      }
```
Bert-R: For readability, please refactor this so the try/catch ends up in a separate private method, so the body of the for-loop becomes something like:
offsetsByGroup.put(groupId, determineOffset(result, groupId));

Bert-R (changes requested):

Bert-R (approved):

Bert-R:
@c4181 Thanks!

